PR #1, opened 2021-10-24
Fix warnings

timo-reichl:
Hi, I found out about this project today and thought I'd fix all the warnings I got while building, as well as some bugs I found using the app.


kolaente:
It's more likely an issue with the setup. Will investigate.

PR #10, opened 2022-08-04
Kanban List View

k9withabone:
This adds a kanban view for lists that looks similar to and acts like the Vikunja web app.


lib/components/BucketTaskCard.dart
```diff
@@ -0,0 +1,326 @@
...
+          child: Text(
+            widget.task.title,
+            style: theme.textTheme.titleMedium.copyWith(
+              color: widget.task.textColor,
+            ),
+          ),
+        ),
+      ],
+    );
+    if (widget.task.dueDate.year > 2) {
```
kolaente: Why the `> 2`? Ar you checking if there's a due date?

lib/api/bucket_implementation.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  @override
+  Future delete(int listId, int bucketId) {
+    return client
+        .delete('/lists/$listId/buckets/$bucketId');
+  }
+
+  @override
+  Future<Bucket> get(int listId, int bucketId) {
+    // Might not exist in the API, it isn't in the docs
```
kolaente: That's correct, the endpoint to get a single bucket is not implemented. Might not hurt to have the method in here but it won't work.

lib/components/BucketTaskCard.dart
```diff
@@ -0,0 +1,326 @@
...
+        label: Text(label.title),
+        labelStyle: theme.textTheme.labelLarge.copyWith(
+          color: label.textColor,
+        ),
+        backgroundColor: label.color,
+      ));
+    });
+    if (widget.task.description.isNotEmpty) {
+      final uncompletedTaskCount = '* [ ]'.allMatches(widget.task.description).length;
+      final completedTaskCount = '* [x]'.allMatches(widget.task.description).length;
```
kolaente: This won't work when the list uses `-` instead of `*`. 
The frontend uses a regex for this: https://kolaente.dev/vikunja/frontend/src/branch/main/src/helpers/checklistFromText.ts

It probably also makes sense to put this in a separate helper function so that we can unit test it on its own.

lib/models/task.dart
```diff
@@ -1,27 +1,33 @@
...
   bool done;
+  Color color;
+  double kanbanPosition;
   User createdBy;
   Duration repeatAfter;
   List<Task> subtasks;
   List<Label> labels;
+  List<TaskAttachment> attachments;
   bool loading = false;
+  // TODO: add position(?)
```
kolaente, line 24: The position is only used in list views, it's fine to add this later.

lib/stores/list_store.dart
```diff
@@ -48,6 +67,26 @@ class ListProvider with ChangeNotifier {
...
+    _isLoading = true;
+    notifyListeners();
+
+    Map<String, List<String>> queryParams = {
+      "page": [page.toString()]
+    };
+
+    return VikunjaGlobal.of(context).bucketService.getAllByList(listId, queryParams).then((response) {
+      if (response.headers["x-pagination-total-pages"] != null) {
+        _maxPages = int.parse(response.headers["x-pagination-total-pages"]);
```
kolaente, line 80: Shouldn't this be handled in the service?

lib/stores/list_store.dart
```diff
@@ -69,32 +108,123 @@ class ListProvider with ChangeNotifier {
...
+      kanbanPosition = (_buckets[newBucketIndex].tasks[index - 1].kanbanPosition
+          + _buckets[newBucketIndex].tasks[index + 1].kanbanPosition) / 2.0;
+
+    task = await VikunjaGlobal.of(context).taskService.update(task.copyWith(
+      bucketId: newBucketId ?? task.bucketId,
+      kanbanPosition: kanbanPosition,
+    ));
+    _buckets[newBucketIndex].tasks[index] = task;
+
+    // make sure first 2 tasks don't have 0 kanbanPosition
```
kolaente: I only figured this out recently, awesome to see you paid attention to this!

lib/stores/list_store.dart
```diff
@@ -69,32 +108,123 @@ class ListProvider with ChangeNotifier {
...
+  
+  Future<void> deleteBucket({BuildContext context, int listId, int bucketId}) {
+    return VikunjaGlobal.of(context).bucketService.delete(listId, bucketId)
+        .then((_) {
+          _buckets.removeWhere((bucket) => bucket.id == bucketId);
+          notifyListeners();
+        });
+  }
+
+  Future<void> moveTaskToBucket({BuildContext context, Task task, int newBucketId, int index}) async {
```
kolaente, line 174: Again, maybe worth moving the position calculation to a helper to make it testable on its own?

lib/theme/theme.dart
```diff
@@ -10,11 +12,19 @@ ThemeData _buildVikunjaTheme(ThemeData base) {
...
+    ),
+    floatingActionButtonTheme: base.floatingActionButtonTheme.copyWith(
+      foregroundColor: vWhite,
+    ),
     buttonTheme: base.buttonTheme.copyWith(
       buttonColor: vPrimary,
       textTheme: ButtonTextTheme.normal,
       colorScheme: base.buttonTheme.colorScheme.copyWith(
         // Why does this not work?
+        // ButtonTheme seems to be obsolete see: https://api.flutter.dev/flutter/material/ButtonThemeData-class.html
```
kolaente: So we could remove this?

kolaente (commented):
Really great work!

My flutter knowledge is rather limited, therefore my comments are more general and not very flutter specific.

kolaente:
Any idea why the pipeline fails?

lib/components/BucketTaskCard.dart
k9withabone: Yes, this was a copy from elsewhere in the app (I believe it was the TaskTile?). I wasn't sure how the frontend does this. This check should probably be done in the Task model.

lib/api/bucket_implementation.dart
k9withabone: I shall comment it out then for potential future use.

lib/models/task.dart
k9withabone: This seems to be a pretty new feature in Vikunja? My build from source as of 7/31 let's you rearrange the tasks but doesn't actually save the positions. But I did just try it on the demo version and it worked. Anyway that's something to add in the list view.

lib/stores/list_store.dart
k9withabone: Yeah it took me awhile to get the task moving right. I actually ended up looking at the frontend code to see how you did it.

lib/stores/list_store.dart
k9withabone: Definitely a good idea.

lib/theme/theme.dart
k9withabone: I believe so, changing the values inside doesn't seem to change anything.

lib/models/task.dart
kolaente: It's a few months old actually. Are you using a recent frontend + api build?

lib/stores/list_store.dart
k9withabone: Probably, honestly when I wrote this I was just copying from `loadTasks` above. Now that I'm looking at it again though, I just realized it clears the existing buckets when it goes to load the next page which is not good. @Benimautner and I were talking about changing the state management stuff anyway, but that will have to be a future PR.

lib/models/task.dart
k9withabone: Yes, both built and updated last Sunday. Using sqlite db, running directly in a Ubuntu 22.04 lxc container on a Proxmox host.

lib/components/BucketTaskCard.dart
k9withabone: Done, moved to helper functions similar to the frontend implementation in `checkboxes_in_text.dart` which are called from methods in `task.dart` which `BucketTaskCard.dart` now uses.

lib/components/BucketTaskCard.dart
k9withabone: Moved the check to `task.dart`, also made it a little more like the frontend in that it does `dueDate.year != 1`.

lib/stores/list_store.dart
k9withabone: Done, I also made the bucket position update use it as well. `calculate_item_position.dart` is very similar to the frontend's `calculateItemPosition.ts`.

lib/stores/list_store.dart
Benimautner: I don't know why pagination is handled there rather than in the service. As @k9withabone said, we should restructure the state management and handle pagination in the service.
I will open an issue regarding state management soon so we can discuss possible options for the rework.

Benimautner:
@kolaente pretty sure the pipeline fails because it cannot access the GitHub Actions secret ANDROID_KEYSTORE_BASE64, which is used to sign the apk. Not sure how to fix this.
I do not have access to the repository's secrets, but citing the settings/secrets page: 'Secrets are not passed to workflows that are triggered by a pull request from a fork.'
Not sure how to fix this.

kolaente:
@Benimautner shouldn't we be able to build the app in a pipeline without signing it? The pipeline is just smoke-testing anyway to see if it still builds.

Benimautner:
@kolaente I updated the pipeline to only sign builds on push. Pull requests are getting built without signing. This does not change the checks for this pull request though.

lib/models/label.dart
```diff
@@ -39,4 +40,6 @@ class Label {
         'updated': updated?.toUtc()?.toIso8601String(),
         'created': created?.toUtc()?.toIso8601String(),
       };
+
+  Color get textColor => color.computeLuminance() > 0.5 ? vLabelDark : vLabelLight;
```
Benimautner: I needed to add a null check:
`  Color get textColor => color != null && color.computeLuminance() > 0.5 ? vLabelDark : vLabelLight;`
Otherwise, I get a 'computeLuminance was called on null' exception

lib/pages/list/list.dart
```diff
@@ -145,60 +581,148 @@ class _ListPageState extends State<ListPage> {
...
+          children: <Widget>[
+            Text('${bucket.title} was deleted.'),
+            Icon(Icons.delete),
+          ],
+        ),
+      ));
+    });
+    _onViewTapped(1);
+  }
 }
```
Benimautner, line 728: Is it possible to move all of the bucket helper functions into their own file? There isn't a reason for such large files, they just get less readable

lib/models/label.dart
k9withabone: What dart version are you using? Could have something to do with null safety checks.

lib/pages/list/list.dart
k9withabone: I think to do it properly would require getting state management sorted out in conjunction. All of this was just to get everything to a working point. Also the project's directory/file structure doesn't lend itself to widgets being broken up easily. What I want to do is break out the list and bucket views from the page and have them handle their own state where possible.

Benimautner: That's fair enough, I think the code structure needs a big overhaul in general.

lib/models/label.dart
Benimautner: Could be, but I don't think so because if I change to a minimum dart version with sound null safety (eg 2.12.0) the app isn't buildable. To upgrade to null safety will be a big undertaking.

Benimautner:
I think this PR is ready to merge, expecially when you take a look at the current state of the codebase in general. Just react to this comment, @kolaente & @k9withabone so I can see whether you think it's ready to merge as well.

lib/models/label.dart
k9withabone: It works on my machine<sup>TM</sup> but 🤷. I am using dart version 2.17.6 for reference. Regardless, I added the null check for now anyway. At some point I do want to do the sound null safety upgrade. It is possible to do it in parts so it could be done over time. The Flutter team has a good [video](https://youtu.be/bvq7wbn4VAA) on null safety migration.

kolaente (approved):

kolaente:
Fine to merge from my point of view!

PR #14, opened 2022-08-23
Fix 'Expected type after 'as''

Szymek887:


Szymek887:
When there was a space between 'as' and '?' app won't build giving error 'Expected type after 'as''. After removing this space it started building

Benimautner:
Can't verify IOS build as I have no IOS device to test it out, but docs say you're right.
Does it work when you build it?

Szymek887:
Yes, when I build it, it works on iOS Simulator

Benimautner:
Amazing, thanks!

PR #16, opened 2022-09-03
null-safety & some other cleanup

k9withabone:


lib/api/client.dart
```diff
@@ -16,13 +16,13 @@ class Client {
   GlobalKey<ScaffoldMessengerState> global;
   final JsonDecoder _decoder = new JsonDecoder();
   final JsonEncoder _encoder = new JsonEncoder();
-  String? _token;
-  String? _base;
+  String _token = '';
+  String _base = '';
```
k9withabone, line 20: This and other changes like this (`String?` --> `String`) were made because a nullable String, when interpolated, prints null instead of an empty string. Ex: `String? string; '$string' == 'null'`

lib/models/bucket.dart
```diff
@@ -7,55 +6,58 @@ import 'package:vikunja_app/models/user.dart';
...
-
-  List<Task> tasks = [];
+    DateTime? created,
+    DateTime? updated,
+    required this.createdBy,
+    List<Task>? tasks,
+  }) {
+    this.created = created ?? DateTime.now();
+    this.updated = created ?? DateTime.now();
+    this.tasks = tasks ?? [];
```
k9withabone: `const` lists are not growable, where `final` lists are. `tasks` is currently manipulated directly when moving tasks between buckets, so it can't be `const`.

lib/models/bucket.dart
```diff
@@ -7,55 +6,58 @@ import 'package:vikunja_app/models/user.dart';
...
+        createdBy = User.fromJson(json['created_by']),
+        tasks = json['tasks'] == null
+            ? []
+            : (json['tasks'] as List<dynamic>)
+                .map((task) => Task.fromJson(task))
+                .toList();
 
   toJSON() => {
-    'id': id,
+    'id': id != -1 ? id : null,
```
k9withabone, line 50: Did this this way, here and in other models, in order to make `id` non-nullable while also allowing new items to have a null id when the request is made to the server to create it. This allows the server to assign a new id.

lib/models/list.dart
```diff
@@ -1,50 +1,53 @@
...
-      "created": this.created?.toIso8601String(),
-      "updated": this.updated?.toIso8601String(),
-      "namespace_id": this.namespaceId
+      'id': id != -1 ? id : null,
+      'title': title,
+      'description': description,
+      'owner': owner.toJSON(),
+      'created': created.toIso8601String(),
+      'updated': updated.toIso8601String(),
+      'namespace_id': namespaceId
```
k9withabone, line 48: Single quotes to make it more consistent with other models.

lib/models/task.dart
```diff
@@ -1,57 +1,63 @@
...
-    this.reminderDates,
+    this.reminderDates = const [],
     this.dueDate,
     this.startDate,
     this.endDate,
     this.parentTaskId,
     this.priority,
     this.repeatAfter,
-    this.color = vBlue, // TODO: decide on color
+    this.color,
```
k9withabone, line 47: Made `color` nullable again (here and for Label). If it's non-nullable then every task that is edited will have the color that is set here have that color set for the task on the server. @Benimautner if you don't like the default look of the task cards in the kanban view that probably should be changed on the UI side, not here.

lib/models/task.dart
```diff
@@ -160,27 +158,27 @@ class Task {
...
-      subtasks: subtasks,
-      labels: labels,
-      attachments: attachments,
+      color: (resetColor ?? false) ? null : (color ?? this.color),
+      kanbanPosition: kanbanPosition ?? this.kanbanPosition,
+      createdBy: createdBy ?? this.createdBy,
+      repeatAfter: repeatAfter ?? this.repeatAfter,
+      subtasks: subtasks ?? this.subtasks,
+      labels: labels ?? this.labels,
+      attachments: attachments ?? this.attachments,
```
k9withabone, line 179: The idea of the `copyWith` method is that it returns a copy of the Task with the modifications specified.

lib/pages/list/list.dart
```diff
@@ -159,7 +159,7 @@ class _ListPageState extends State<ListPage> {
   ListView _listView(BuildContext context) {
     return ListView.builder(
       padding: EdgeInsets.symmetric(vertical: 8.0),
-      itemCount: taskState!.tasks.length,
+      itemCount: taskState!.tasks.length * 2,
```
k9withabone, line 162: `* 2` to account for the Dividers.

lib/pages/list/task_edit.dart
```diff
@@ -221,15 +220,15 @@ class _TaskEditPageState extends State<TaskEditPage> {
                         ),
                         onTap: () {
                           // We add a new entry every time we add a new input, to make sure all inputs have a place where they can put their value.
-                          _reminderDates!.add(null);
-                          var currentIndex = _reminderDates!.length - 1;
+                          _reminderDates.add(DateTime(0));
```
k9withabone, line 223: Adding a `null` threw an exception even if the list items were nullable.

lib/pages/user/login.dart
```diff
@@ -131,7 +131,7 @@ class _LoginPageState extends State<LoginPage> {
                                         _loginUser(context);
                                       }
                                     }
-                                  : () => null,
+                                  : null,
```
k9withabone, line 134: Having a null `onPressed` shows the button as disabled which I think was the intention.

k9withabone:
I think the check might be failing because there is a new version of Flutter available and it's trying to use the newer versions which are incompatible with the versions in the pubspec.lock maybe?

lib/api/client.dart
Benimautner: Ah, makes sense. I was wondering why you would change that!

lib/pages/list/task_edit.dart
Benimautner: Weird. It worked for me. But this is cleaner

lib/models/task.dart
Benimautner: That was something I wanted to fix, because the way it looked after I changed that was horrible. Just haven't gotten around to it. Thanks!

lib/models/bucket.dart
Benimautner: So why not make the id nullable in the first place?

lib/pages/list/task_edit.dart
```diff
@@ -524,11 +530,11 @@ class _TaskEditPageState extends State<TaskEditPage> {
         ),
         actions: <TextButton>[
           TextButton(
-            child: Text('Cancel'),
+            child: Text('CANCEL'),
             onPressed: () => Navigator.of(context).pop(),
           ),
           TextButton(
-            child: Text('Reset'),
+            child: Text('RESET'),
```
Benimautner, line 537: Why did you capitalize that?

lib/pages/namespace/namespace_edit.dart
```diff
@@ -97,10 +104,13 @@ class _NamespaceEditPageState extends State<NamespaceEditPage> {
...
-        title: _name,
-        description: _description,
-        owner: widget.namespace.owner);
+      id: widget.namespace.id,
+      title: _name,
+      description: _description,
+      owner: widget.namespace.owner,
+      created: widget.namespace.created,
+      updated: widget.namespace.updated,
+    );
```
Benimautner: We should either make title and description of Namespace non-final or add a copyWidth function.

lib/utils/validator.dart
```diff
@@ -1,13 +1,13 @@
 final RegExp _emailRegex = new RegExp(
     r'^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$');
 
-bool isEmail(email) {
-  return _emailRegex.hasMatch(email);
+bool isEmail(String? email) {
+  return _emailRegex.hasMatch(email ?? '');
```
Benimautner: You could return false whenever email is null. Would avoid the regex call.

lib/utils/validator.dart
```diff
@@ -1,13 +1,13 @@
...
 }
 
 final RegExp _url = new RegExp(
     r'https?:\/\/((([a-zA-Z0-9.\-\_]+)\.[a-zA-Z]+)|(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}))(:[0-9]+)?');
 
-bool isUrl(url) {
-  return _url.hasMatch(url);
+bool isUrl(String? url) {
+  return _url.hasMatch(url ?? '');
```
Benimautner: See previous comment.

lib/models/bucket.dart
Benimautner: So why not make the id nullable in the first place?

Benimautner (commented):
In general I'm okay with all the changes. Also, I'm sorry, some changes I made in the migrations were not well though through, but my schedule has been very full in the last couple of weeks.

lib/models/bucket.dart
k9withabone: Because then we have to do null checks everywhere we want to use id, which is a lot of places. Also, you had it has non-nullable, but required and setting it to 0 when creating a new item. This just formalizes that with a default value instead. However, it does make sense semantically for id to be nullable to represent an item that hasn't had an id assigned to it yet. So, I'm not sure what we should do. I'm not sure what happens when give the Vikunja API a id <= 0, but I sort-of assumed that it would return an error.

Eventually, when we get around to adding offline capability, we will have to look at how the API calculates the id and do that too, so that items can be added to newly created items while offline and then do some sort of migration when coming back online. That's going to be really complicated.

lib/pages/list/task_edit.dart
k9withabone: Following Material design guidelines. See: https://material.io/components/buttons#text-button

lib/pages/namespace/namespace_edit.dart
k9withabone: Agreed. I'll do the copyWith function for the moment, but I think we need to have a larger discussion on whether we want to have mutable or immutable models. I am personally leaning towards immutable as that makes the most sense to me especially with Flutter widgets being immutable. However, neither Dart nor Flutter are opinionated on the subject for data models.

lib/utils/validator.dart
k9withabone: That makes sense.

lib/models/bucket.dart
kolaente: In Go the "null" value of an integer is `0`. That's what the api is doing.

k9withabone: Oh, ok. What happens when you give it a task with a `0` listId or similar things for other items?

kolaente: Not 100% sure but you'll either get an error or some kind of undefined behaviour where the task will be created in the db but you won't be able to access it as it does not belong to a list.

kolaente: Just checked: When you try to set a `list_id` of a task to 0 it will be overridden with the actual id of that task.

lib/pages/landing_page.dart
```diff
@@ -103,20 +103,31 @@ class LandingPageState extends State<LandingPage> with AfterLayoutMixin<LandingP
...
+      return;
+    }
+
+    await globalState.taskService.add(
+      defaultList!,
+      Task(
+        createdBy: globalState.currentUser!,
+        listId: defaultList!,
+      ),
+    );
```
Benimautner: You're missing title and due date here. Adding tasks from the landing page does not work.

lib/models/list.dart
```diff
@@ -1,50 +1,51 @@
...
-      "owner": this.owner?.toJSON(),
-      "created": this.created?.toIso8601String(),
-      "updated": this.updated?.toIso8601String(),
-      "namespace_id": this.namespaceId
+      'id': id != -1 ? id : null,
+      'title': title,
+      'description': description,
+      'owner': owner.toJSON(),
+      'created': created.toIso8601String(),
+      'updated': updated.toIso8601String(),
```
Benimautner: .toUtc is needed here and in every other toJSON as the API fails to handle timestamps with timezones correctly. I'm not sure if this is the API's fault or flutter's. toUtc is a workaround, we should probably address this at some point in the future.

lib/components/SliverBucketList.dart
```diff
@@ -33,14 +32,16 @@ class SliverBucketList extends StatelessWidget {
...
+    await Provider.of<ListProvider>(context, listen: false).moveTaskToBucket(
       context: context,
       task: task,
       newBucketId: bucket.id,
       index: index,
-    ).then((_) => ScaffoldMessenger.of(context).showSnackBar(SnackBar(
+    );
+
+    ScaffoldMessenger.of(context).showSnackBar(SnackBar(
       content: Text('${task.title} was moved to ${bucket.title} successfully!'),
```
Benimautner: Please add quotes around the task title and bucket title.

lib/models/bucket.dart
Benimautner: > Eventually, when we get around to adding offline capability, we will have to look at how the API calculates the id and do that too, so that items can be added to newly created items while offline and then do some sort of migration when coming back online. That's going to be really complicated.

If we ever wanted to add offline capabilities, I'd just assign sequentially new ids and set a flag on the task with 'is_local_id' and when you come back online you just go over all tasks with that flag set, ignore the local id and get a server id. However, we need better state management and a bit of cleanup before we should think about offline capabilities.


I'm fine with how you're leaving 0 for the id as an indicator that it's new to avoid all of the null checks throughout the code, but I think logically it would make more sense to set it to null.

Benimautner (changes requested):
If we iron out those last few things I think this is ready to merge!

lib/models/bucket.dart
k9withabone: So should just change the default id to 0 or should I make it nullable and add the null checks?

Benimautner: Make it default to 0 and leave it non-nullable.

lib/components/SliverBucketList.dart
k9withabone: Done, I also add them in other user facing places.

lib/models/list.dart
k9withabone: Hmm, seems to work for me. When I made requests directly (in python) it was giving me the proper timezone and it didn't seem to have difficulty when I gave it back. I added the `.toUtc()` anyway.
Regardless, I'm not sure these are editable for the client as the documentation for `'updated'` says:

> updated	
> string
> A timestamp when this task was last updated. You cannot change this value.

So perhaps we should remove the `'created'` and `'updated'` from the `toJSON`s?

lib/pages/landing_page.dart
k9withabone: Whoops, fixed.

Benimautner:
Thanks so much for your work! Merged it and will release a new version soon

kolaente:
Nice work you two!

lib/models/bucket.dart
k9withabone: Done.

k9withabone:
Don't merge it into main quite yet. Just realized there is a couple other quirks, such as not being able to remove due dates. I'll submit another PR shortly.

Benimautner:
Sorry, read your comment too late. Create a new PR and I'll merge it.

PR #17, opened 2022-09-07
Null safety fixes

k9withabone:
I think I got all the bugs? If there is more let me know and I'll fix them.


lib/models/bucket.dart
```diff
@@ -14,7 +14,7 @@ class Bucket {
   final List<Task> tasks;
 
   Bucket({
-    this.id = -1,
+    this.id = 0,
```
k9withabone, line 17: Changed to 0 as discussed in #16.

lib/models/task.dart
```diff
@@ -138,8 +138,11 @@ class Task {
     DateTime? created,
     DateTime? updated,
     DateTime? dueDate,
+    bool? clearDueDate,
     DateTime? startDate,
+    bool? clearStartDate,
     DateTime? endDate,
+    bool? clearEndDate,
```
k9withabone: This allows for clearing the dates by setting them to null if the corresponding bool is true.

lib/utils/repeat_after_parse.dart
```diff
@@ -43,13 +43,16 @@ getRepeatAfterValueFromDuration(Duration? repeatAfter) {
   return repeatAfter.inHours;
 }
 
-getDurationFromType(String? value, String? type) {
+Duration? getDurationFromType(String? value, String? type) {
```
k9withabone, line 46: Added return types to make it less ambiguous.

lib/pages/list/task_edit.dart
```diff
@@ -370,8 +370,11 @@ class _TaskEditPageState extends State<TaskEditPage> {
       description: _description,
       reminderDates: _reminderDates,
       dueDate: _dueDate,
+      clearDueDate: _dueDate == null,
       startDate: _startDate,
+      clearStartDate: _startDate == null,
       endDate: _endDate,
+      clearEndDate: _endDate == null,
```
k9withabone: The dates will be null when cleared.

lib/models/task.dart
Benimautner: This, in theory, makes sense, but is also why I hate the copyWith function. You can't differentiate between variables set to null on purpose and those that should not get set. But changing that requires a larger rewrite.
Regardless, would you mind creating a class that stores the nullable datetime and the clear status in one object? I'm not a big fan of having those bools clutter up the class. Thanks!

Something along the lines of:
`class ClearableDateTime {
  DateTime? date;
  bool clear = false;
}
`

k9withabone: Just thought of another solution. How about setting their default values and checking for that? Like so:
```
DateTime? dueDate = Duration.zero,
DateTime? startDate = Duration.zero,
DateTime? endDate = Duration.zero,
```
and then checking:
```
dueDate = (dueDate == Duration.zero) ? null : (dueDate ?? this.dueDate);
```

Benimautner: This would not represent exactly what we want to store. Also, I think you cannot assign Duration to DateTime, so the nearest equivalent would be `DateTime(0)`, but then there's no differentiating between the 1st of January 0000 and no date. Probably does not matter as long as we don't actually send that to the API though. Still, I'd go with an extra class.

k9withabone: Sorry I meant `DateTime(0)`.

k9withabone: How about moving the resets to their own class? That way we don't have to change the underlying class.
```
class TaskResets {
  final bool resetDueDate, resetStartDate, resetEndDate, resetColor;

  const TaskResets({
    this.resetDueDate = false,
    this.resetStartDate = false,
    this.resetEndDate = false,
    this.resetColor = false,
  });
}
```

Benimautner: That would work, but I still think a class for clearable datetimes would make more sense.
```
class ClearableDateTime {
  DateTime? _date;
  bool clear = false;

  ClearableDateTime(DateTime? _newdate) {
    _date = _newdate;
    clear = _newdate == null;
  }
 
  DateTime? get date {
    return clear ? _date : null;
  }
}
```

Then change every field from DateTime? to ClearableDateTime. In the copyWith call, you could then use `dueDate = ClearableDateTime(_dueDate)`, and in the toJSON function `'due_date': dueDate.date.toUtc().toIso8601String()`.
That way we could avoid this: `dueDate: (clearDueDate ?? false) ? null : (dueDate ?? this.dueDate),`.

Disclaimer: I'm not 100% sure this makes sense.

k9withabone: I was sort-of waiting until the immutable vs mutable models discussion was had but another option is to use the [freezed package](https://pub.dev/packages/freezed) which supports setting things to null in its `copyWith`. We could also combine it with [json_serializable](https://pub.dev/packages/json_serializable) for the `toJson` and `fromJson`.

Benimautner: Well as I said I am not a big fan of working with `copyWith`. In general, I'd be in favor of changing models to be mutable. But if we do that, we also have to finally talk about state management.
But that's not the scope of this PR, we should talk about mutable/immutable in a separate issue/PR.
After having a quick look, freezed seems great. I'm also not sure why we're not making more use of the factories provided by json_serializable.

k9withabone: So for now I made the dates mutable so that they can be directly set to null.

Benimautner: This looks much cleaner. Thanks! Do you think it's ready to merge? It looks good to me.

k9withabone: Yeah I think it's ready to merge.

PR #68, opened 2024-04-03
Various improvements

denysvitali:
This is a huge PR (basically syncing my changes with yours), so let me know if you'd rather have me split it to make things easier.

My `main` branch features the following:

- Fix #66: It is now possible to use one of the User CAs to securely connect to the backend when a CA is not trusted by Flutter (but it's trusted by the system)
- Workaround for #67: add support for `X-Client-Token`, a special header that will be added to all the requests. To set this parameter, tap quickly 5 times the Vikunja logo in the login page to show the field
- Improve Landing Page view by adding Material [Chips](https://m3.material.io/components/chips/overview) to show the project associated with the to-do item
- Strike-through the text of completed To-Do items
- Show a description / due date for the To-Do item
- ... other things that I don't remember :D

PR #69, opened 2024-04-05
Update dependencies

denysvitali:
This PR updates the dependencies and is part of the split of #68.
A lot of code style changes here are due to the auto-formatter of Flutter (why isn't the project using it?)

**This PR depends on #71**, review only 73624b2870ad06e7b622ac0b7d8e24750c1c616c until #71 is merged


kolaente:
Can you rebase this on latest main?

Benimautner:
Re: why the project wasn't using flutter format: Basically, we just never started haha. It's definitely a good idea. As @kolaente said, please rebase all your PRs on latest main so the changelog doesn't include all the formatting.

And I don't want to push the minsdk all the way up to Android 13, so if there's any way you can downgrade the dependencies, that would be amazing!

PR #70, opened 2024-04-05
Add web support

denysvitali:
Depends on #69 (stacked PR, the diff is going to be huge until #69 is merged, just review 5b82a47)

This PR adds support for the `web` platform to make it easier to develop the app further. Eventually, this could be a second frontend to Vikunja - although this is not really the reason why I'm doing this.

With this PR, I've also fixed an issue that was seen during Hot Reload in the web, where the theme was not correctly reset.


Benimautner:
I think it's fine to have it, but since Vikunja already has a great web frontend with a lot more features, I don't really see a reason to add (and then maintain) this.

denysvitali:
For me the biggest advantage is to be able to develop the features locally with Chrome (or Chromium).

We can maybe keep this target enabled but without officially supporting it

Benimautner:
What do you mean by developing locally with chrome? The official frontend is in Vue, you can develop that there, too, no?

We could, but it's just not what I would use flutter for for vikunja. @kolaente let me know what you think pls

kolaente:
Similar to what Beni said, we won't use the flutter web app as a replacement for the current web frontend anytime soon. I think having an easy way to develop for it without requiring a full-blown Android dev environment is a very good thing.

Can you add a note to the readme about how to get started with flutter web and how we'll only use it for development?

Benimautner:
The Readme is on my todo list anyways, so I'ma just add this when I update it!

PR #71, opened 2024-04-05
Dart Format

denysvitali:
Since most of my PRs look like a lot has changed (because this project doesn't follow the idiomatic Dart format), this PR is the result of running:

```
dart format .
```

From this PR forward, all the changes should be in the idiomatic Dart format - so that the changes won't end up looking like a huge mess.


denysvitali:
Sure thing. Let me do that!

This is indeed the first PR that needs to be merged - otherwise you'll see in the other PRs (unless you check commit by commit) huge, unrelated changes

lib/api/list_implementation.dart
```diff
@@ -32,60 +34,55 @@ class ListAPIService extends APIService implements ListService {
...
+    return client.get('/lists').then((list) {
+      if (list == null || list.statusCode != 200) return null;
+      if (list.body.toString().isEmpty) return Future.value([]);
+      print(list.statusCode);
+      return convertList(list.body, (result) => TaskList.fromJson(result));
+    });
   }
 
   @override
   Future<List<TaskList>?> getByNamespace(int namespaceId) {
```
kolaente, line 57: @Benimautner Is this still used? The api endpoints don't work anymore.

Benimautner: No, there's a bunch of dead code for editing lists & namespaces which should be cleaned. It's still in there bc I wanted to check whether I could adapt the pages to edit projects.

Benimautner (commented):
looks good. I'm looking forward to this happening as a workflow!

PR #79, opened 2024-06-06
feat: build app in CI

kolaente:


kolaente:
Not sure why the build fails all of a sudden, but it's reproducible locally. Will look into a fix.

Benimautner:
lgtm!

PR #107, opened 2025-02-15
feat: upgrade project to work with the latest version of flutter

### LINKED ISSUE #106: Cannot build project
There are a few features I would like to implement and as I have some familiarity with flutter I wanted to give it a go. However I have as yet been unable to build it as there are so many depreciacions and errors with the various libraries and settings used here.

Can I find out what versions of flutter and dart are used to build this? And is there any plan to update all the dependencies?

brokoli18:
There are a few changes here all of which go towards making the project work with the latest version of flutter and compile properly. The summery of them are below

* Upgrade/Downgrade versions of kotlin, jvm, gradle etc... to enable the app to compile
* Run a `flutter pub upgrade --major versions` to upgrade dependencies in `pubspec.yaml`
    * We are running afowl of [v1 flutter binding depreciation](https://docs.flutter.dev/release/breaking-changes/android-v1-embedding-create-deprecation), so a lot of the versions of dependencies being used are broken. Most are fixed with an upgrade, however inappwebview and workmanager need the latest code to be pulled from git
    * After upgrading packages some options are no longer required and some are now mandatory. This only affected `notifications.yaml` and the file was updated accordingly

Flutter version:
```
Flutter 3.24.4 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 603104015d (4 months ago) • 2024-10-24 08:01:25 -0700
Engine • revision db49896cf2
Tools • Dart 3.5.4 • DevTools 2.37.3
```

Dart version:
```
Dart SDK version: 3.5.4 (stable) (Wed Oct 16 16:18:51 2024 +0000) on "linux_x64"
```

Resolves #106

### DIFF
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -42,5 +42,5 @@ jobs:
           channel: stable
       - name: Check Dart Format
         run: |
-          fluter pub get
+          flutter pub get
           dart format --set-exit-if-changed .
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -41,4 +41,6 @@ jobs:
         with:
           channel: stable
       - name: Check Dart Format
-        run: dart format --set-exit-if-changed .
+        run: |
+          fluter pub get
+          dart format --set-exit-if-changed .
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -44,6 +44,9 @@ android {
         targetCompatibility = JavaVersion.VERSION_1_8
     }
 
+    kotlinOptions {
+        jvmTarget = JavaVersion.VERSION_1_8
+    }
 
     defaultConfig {
         applicationId "io.vikunja.app"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -5,9 +5,6 @@ allprojects {
     }
 }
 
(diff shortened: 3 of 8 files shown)


kolaente:
Looks pretty good already! Can you fix the lint errors?

brokoli18:
> Looks pretty good already! Can you fix the lint errors?

I am seing some fairly strange behaavior where when I run `dart format` on my local machine it changes only 1 fine, but when I run the pipeline it changes 50. Dart and flutter versions are the same across both so i am not sure why its not working. Perhaps there is an `analysis_options.yaml` file that is downloaded on the runner? @Benimautner Have you ran into this before?

brokoli18:
I have figured it out. `dart format` will behave differently depending on if you have pulled all the packages or not, the pipeline currently does not pull any packages and so the formatting fails.

@kolaente I think it may be a better idea to put a step into the pipeline to do a pub get before running the format so the packages are formatted properly. What do you think?

brokoli18:
I took the liberty of tweaking the pipeline slightly to get packages before performing a format. Now it works without a problem.

kolaente (approved):
Thanks!

Benimautner:
thanks both!

PR #109, opened 2025-02-25
Add a filter to show only tasks due today to main screen

brokoli18:
When opening the app I usually want to see tasks that I need to do today at a quick glance. I added a filter (that is on by default) that shows only tasks due today.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,6 +94,7 @@ app.*.map.json
 /android/app/debug
 /android/app/profile
 /android/app/release
+/android/app/.cxx
 
 # VS Code
 .vscode/settings.json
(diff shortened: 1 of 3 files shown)


Benimautner:
Not a huge fan of this since Vikunja supports landing page filters anyways, which the app respects, and you can just create a filter for that, offering more flexibility.
image

brokoli18:
In that case is there a particular use case the `Only show tasks with due date` button is doing that cant be achieved using the same filter?

Benimautner:
That's only there for legacy reasons (I implemented it before I knew of the filter functionality/maybe before that existed? I don't remember) - regardless, we could remove that as well.

brokoli18:
Ah yes I see you did the same thing I did in this MR :grin:. I will close this and use filters as you suggested

PR #122, opened 2025-06-23
fix(ci): flutter build

kolaente:


kolaente:
The build works, this is redundant.

PR #125, opened 2025-08-01
Fixed endless rebuild

leschbenedikt:
Hi there,

this is my first PR for this project. Plan todo more in the future as I started to use vikunja and like to have an app.

- Code cleanup
- Replaced workmanager dependency
- Fixed endless rebuild caused by theme notification

PR #126, opened 2025-08-02
Fixed endless rebuild

leschbenedikt:
Hi, 

this is the replacement PR for https://github.com/go-vikunja/app/pull/125

Fixed an endless rebuild of the main app widget.
The ListenableBuilder was triggered by the FutureBuilder cause it sets the theme when loading data.


kolaente (approved):
Thanks!

I went ahead and adjusted the commit messages to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) - it would be awesome if you could use them in the future (but I didn't tell you this beforehand, so the blame here is on me).

PR #128, opened 2025-08-03
chore: replaced deprecated WillPopScope with PopScope

leschbenedikt:
WillPopScope is deprecated as it will not work with predictive back as is replaced by PopScope.

In overview.dart is was removed completely as it was never used


kolaente:
Thanks!

PR #129, opened 2025-08-03
fix: color string in task edit page wrong on some devices

leschbenedikt:
Colorstring in task edit page was displayed weird on some devices.

image

The calculation is base on toString which is flawed after all.
Replaced by a calculation that is already available and should be more reliable.

### DIFF
--- a/lib/pages/project/task_edit.dart
+++ b/lib/pages/project/task_edit.dart
@@ -442,18 +442,14 @@ class _TaskEditPageState extends State<TaskEditPage> {
                           Padding(
                             padding: const EdgeInsets.only(left: 15),
                             child: () {
-                              String? colorString = (_resetColor
-                                      ? null
-                                      : (_color ?? widget.task.color))
-                                  ?.toString();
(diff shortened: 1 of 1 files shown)


lib/pages/project/task_edit.dart
```diff
@@ -442,18 +442,14 @@ class _TaskEditPageState extends State<TaskEditPage> {
...
-                                      ? null
-                                      : (_color ?? widget.task.color))
-                                  ?.toString();
-                              colorString = colorString
-                                  ?.substring(10, colorString.length - 1)
-                                  .toUpperCase();
-                              colorString = colorString != null
-                                  ? '#$colorString'
-                                  : 'None';
+                              Color? colorString = (_resetColor
```
kolaente: Nitpick: If this is not a `String` but a `Color`, should it still be called `colorString`?

leschbenedikt: You're absolutely right. Changed it

PR #132, opened 2025-08-05
feat(ci): release on Google Play via fastlane

kolaente:


kolaente:
Had to rotate the keys in the play store console - the new key is usable from aug 7

PR #134, opened 2025-08-08
refactor: moved files according to new layer-first folder structure &…

leschbenedikt:
… renamed services to data source to match common flutter naming


kolaente:
Unsure why the build failed, did it just run out of memory?

leschbenedikt:
Yeah seems like. Can we increase the memory and try again?

kolaente:
The ci job running this has 16GB which should be plenty. Is this a problem of the runner not having enough or the jvm not using enough?

kolaente:
Increased the memory for gradle via the gradle opts in https://github.com/go-vikunja/app/commit/92e1dc8b22a9071d1dc6fec8eac4e962c9ce2954, let's see if that fixes here

lib/data/data_sources/project_view_data_source.dart
```diff
@@ -1,8 +1,9 @@
-import 'package:vikunja_app/api/service.dart';
-import 'package:vikunja_app/models/view.dart';
-import 'package:vikunja_app/service/services.dart';
+import 'package:vikunja_app/core/network/service.dart';
+import 'package:vikunja_app/data/models/view.dart';
+import 'package:vikunja_app/core/services.dart';
 
-class ProjectViewAPIService extends APIService implements ProjectViewService {
+class ProjectViewAPIService extends RemoteDataSource
+    implements ProjectViewService {
```
kolaente, line 6: wrong line break? Or was it formatted like this by `dart format`?

kolaente: On the other hand, the linter passes so I guess it's fine.

leschbenedikt: Should have been formatted with dart format.

PR #135, opened 2025-08-09
Repositories added

leschbenedikt:
Repositories added and model split into entities and models.
Creates a separation of layer as usual in clean architecture.

Classes are by far not final, Just a step to the final architecture

### DIFF
--- a/lib/domain/entities/task_label_bulk.dart
+++ b/lib/domain/entities/task_label_bulk.dart
@@ -1,7 +1,0 @@
-import 'package:vikunja_app/domain/entities/label.dart';
-
-class LabelTaskBulk {
-  final List<Label> labels;
-
-  LabelTaskBulk({required this.labels});
-}


lib/data/models/bucket_dto.dart
```diff
@@ -1,19 +1,19 @@
 import 'package:json_annotation/json_annotation.dart';
-
-import 'package:vikunja_app/data/models/task.dart';
+import 'package:vikunja_app/data/models/task_dto.dart';
 import 'package:vikunja_app/data/models/user.dart';
+import 'package:vikunja_app/domain/entities/bucket.dart';
 
 @JsonSerializable()
-class Bucket {
+class BucketDto {
```
kolaente, line 7: What's a `Dto`? A data object?

lib/data/data_sources/bucket_data_source.dart
```diff
@@ -1,63 +1,49 @@
...
-  Future<Bucket> get(int listId, int bucketId) {
-    return client
-        .get('/lists/$listId/buckets/$bucketId')
-        .then((response) => Bucket.fromJSON(response.body));
-  }
-  */
-
-  @override
-  Future<Response?> getAllByList(int projectId, int viewId,
+  Future<Response<List<BucketDto>>?> getAllByList(int projectId, int viewId,
```
kolaente, line 24: Unsure, but IIRC this function name is a relic from the time when projects were called lists, so we should be able to rename this to `getAllByProject`.

lib/data/data_sources/label_data_source.dart
```diff
@@ -1,50 +1,44 @@
...
       if (response == null) return null;
-      return Label.fromJson(response.body);
+      return LabelDto.fromJson(response.body);
     });
   }
 
-  @override
-  Future<Label?> get(int labelID) {
+  Future<LabelDto?> get(int labelID) {
     return client.get('/labels/$labelID').then((response) {
```
kolaente, line 23: Out of curiosity: What is the difference between `'/labels/$labelID'` and `'/labels/${labelID}'`?

lib/data/models/task_dto.dart
```diff
@@ -25,34 +26,40 @@ class TaskReminder {
...
+  static TaskReminderDto fromDomain(TaskReminder b) =>
+      TaskReminderDto(b.reminder, b.relative_period, b.relative_to);
 }
 
 @JsonSerializable()
-class Task {
+class TaskDto {
   final int id;
   final int? parentTaskId, priority, bucketId;
   //final int? listId;
```
kolaente, line 41: This can be removed
```suggestion
```

lib/domain/entities/task.dart
```diff
@@ -0,0 +1,148 @@
...
+        'relative_to': relative_to,
+        'reminder': reminder.toUtc().toIso8601String(),
+      };
+}
+
+@JsonSerializable()
+class Task {
+  final int id;
+  final int? parentTaskId, priority, bucketId;
+  //final int? listId;
```
kolaente:
```suggestion
```

lib/domain/entities/task.dart
```diff
@@ -0,0 +1,148 @@
...
+  final bool done;
+  Color? color;
+  final double? position;
+  final double? percent_done;
+  final User createdBy;
+  Duration? repeatAfter;
+  final List<Task> subtasks;
+  final List<Label> labels;
+  final List<TaskAttachment> attachments;
+  // TODO: add position(?)
```
kolaente, line 49: The position depends on the view that is used, but at some point we should add it as well. But it should be shown in the UI as well, so it probably makes sense to put this in a different PR.

lib/domain/entities/task_label_bulk.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:vikunja_app/domain/entities/label.dart';
+
+class LabelTaskBulk {
```
kolaente: Is the bulk label managment used?

kolaente:
Speaking from my lack of Flutter knowledge, what is the reason behind adding abstract classes for the repositories and then implementing them, when they are only implemented once? Mocking in tests?

lib/data/models/bucket_dto.dart
leschbenedikt: It's a data transfer object. Often used for models in the data layer. Backend and Frontend. But other naming schemas also exist. Helps to distinguish between data and domain layer models.

lib/data/data_sources/bucket_data_source.dart
leschbenedikt: Yeah we should question and improve the naming at some places. But for now I just wanted a "copy" of the data layer cause the PR was already this big. But I agree

lib/domain/entities/task_label_bulk.dart
leschbenedikt: You're right, will check that

lib/domain/entities/task.dart
leschbenedikt: This TODO was already there. Didn't check on those for now.

lib/data/data_sources/label_data_source.dart
leschbenedikt: Nothing. You can omit the braces. You only need them if you want to access the variable like ${label.id}

https://dart.dev/language/built-in-types#strings

lib/domain/entities/task_label_bulk.dart
leschbenedikt: Yep, wasn't used only the dto was. Fixed it.

lib/data/models/bucket_dto.dart
kolaente: Gotcha, makes sense!

lib/data/data_sources/label_data_source.dart
kolaente: Thanks!

PR #141, opened 2025-08-18
Introducing riverpod

leschbenedikt:
This is the first PR that adds riverpod.

I refactored the settings page to use controllers that agregate the data and where the UI can watch for changes.

 To prevent confusion the relevant parts in the old settings manager were removed which led to further necessary changes in other components

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -64,7 +64,6 @@ void callbackDispatcher() {
       return nc
           .scheduleDueNotifications(taskService)
           .then((value) => Future.value(true));
-      // });
     } else if (task == "refresh-token") {
       final FlutterSecureStorage _storage = new FlutterSecureStorage();
 


lib/core/di/reppository_provider.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:flutter_riverpod/flutter_riverpod.dart';
```
kolaente: Typo? This file should be called `repository_provider.dart` (single p)

lib/main.dart
```diff
@@ -47,23 +50,21 @@ void callbackDispatcher() {
...
+      print("ignoring: $ignoreCertificates");
+      client.reloadIgnoreCerts(ignoreCertificates);
+
+      TaskRepository taskService = TaskRepositoryImpl(TaskDataSource(client));
+      NotificationClass nc = NotificationClass();
+      await nc.notificationInitializer();
+      return nc
+          .scheduleDueNotifications(taskService)
+          .then((value) => Future.value(true));
+      // });
```
kolaente: This seems like it can be deleted

```suggestion
```

lib/core/di/reppository_provider.dart
leschbenedikt: Thanks, fixed it

PR #142, opened 2025-08-29
refactor: migrated task page to riverpod

leschbenedikt:


kolaente:
Nice work! It seems to work great when I open a project and list all tasks there. However, when I open the home page, I get this:



(Tested with the demo and demo user)

kolaente:
This might be this issue: https://vikunja.sentry.io/issues/6813896786/events/dabe4f703a0845b4bf22f37b8516fa5d/ - but since that first happened a while ago, it might be unrelated to this PR?

leschbenedikt:
Yes seems like an old issue that is now more visible.
Can't even reproduce it on my side on this branch

All the error handling and mapping will be better once the rest is migrated to riverpod.
The app is mostly done on my side but I split it up in multiple PRs.

PR #144, opened 2025-09-02
Refactor/project list page

leschbenedikt:


lib/presentation/pages/project/project_list_page.dart
```diff
@@ -0,0 +1,96 @@
...
+    if (project.subprojects?.isEmpty == true) {
+      return ListTile(
+        leading: Icon(Icons.list),
+        title: Text(project.title),
+        onTap: () {
+          _navigateToProject(ref.context, project);
+        },
+      );
+    } else {
+      return VikunjaExpansionTile(
```
kolaente: nit: this could be an early return (removing the `else`)

lib/presentation/widgets/project/add_project_dialog.dart
```diff
@@ -0,0 +1,50 @@
...
+        Row(children: <Widget>[
+          Expanded(
+            child: new TextField(
+              autofocus: true,
+              decoration: widget.decoration,
+              controller: textController,
+            ),
+          ),
+        ]),
+        //],)
```
kolaente: This can be removed

```suggestion
```

lib/presentation/pages/project/project_list_page.dart
leschbenedikt: Done.
Also initialized the subproject as empty list to avoid nullpointers

PR #146, opened 2025-09-15
refactor: migrated login and auth to riverpod

leschbenedikt:


lib/core/background_work.dart
```diff
@@ -0,0 +1,64 @@
...
+@pragma('vm:entry-point')
+void callbackDispatcher() {
+  if (kIsWeb) {
+    return;
+  }
+  Workmanager().executeTask((task, inputData) async {
+    print(
+      "Native called background task: $task",
+    ); //simpleTask will be emitted here.
+    if (task == "update-tasks") {
```
kolaente: Maybe these ifs would benefit from a swtich case?

lib/core/services.dart
```diff
@@ -69,32 +64,45 @@ class TaskServiceOption<T> {
...
   TaskServiceOption<TaskServiceOptionOrderBy>(
-      "order_by", TaskServiceOptionOrderBy.asc),
-  TaskServiceOption<TaskServiceOptionFilterBy>("filter_by",
-      [TaskServiceOptionFilterBy.done, TaskServiceOptionFilterBy.due_date]),
-  TaskServiceOption<TaskServiceOptionFilterValue>("filter_value",
-      [TaskServiceOptionFilterValue.enum_false, '1970-01-01T00:00:00.000Z']),
+    "order_by",
+    TaskServiceOptionOrderBy.asc,
+  ),
+  TaskServiceOption<TaskServiceOptionFilterBy>("filter_by", [
```
kolaente: These are the old filter options which don't work anymore with the latest versions. Use a single `filter` string and the query language instead.

https://vikunja.io/docs/filters

lib/presentation/pages/login/register_page.dart
```diff
@@ -0,0 +1,166 @@
...
+                ),
+                Builder(
+                  builder: (context) => FancyButton(
+                    onPressed: !_loading
+                        ? () {
+                            if (_formKey.currentState!.validate()) {
+                              _formKey.currentState?.save();
+                              _registerUser(context);
+                            } else {
+                              print("awhat");
```
kolaente: what?

lib/presentation/pages/project/project_edit.dart
```diff
@@ -33,91 +35,91 @@ class ProjectEditPageState extends ConsumerState<ProjectEditPage> {
...
+                child: TextFormField(
+                  maxLines: null,
+                  keyboardType: TextInputType.multiline,
+                  initialValue: widget.project.title,
+                  validator: (title) {
+                    if (title == null) {
+                      return "Title can't be null";
+                    }
+
+                    if (title.length < 3 || title.length > 250) {
```
kolaente: Actually, the limit on the api is at least 1 and at most 250

lib/presentation/widgets/project/kanban/bucket_delete_dialog.dart
```diff
@@ -5,24 +5,23 @@ class TaskDeleteDialog extends ConsumerWidget {
...
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     return AlertDialog(
       title: Text('Delete Bucket'),
-      content: Text('Are you sure you want to delete this bucket?'
-          "You won't delete any tasks, they will be moved to the default column"),
+      content: Text(
+        'Are you sure you want to delete this bucket?'
+        "You won't delete any tasks, they will be moved to the default column",
```
kolaente:
```suggestion
        'You won't delete any tasks, they will be moved to the default bucket',
```

pubspec.yaml
```diff
@@ -4,29 +4,23 @@ description: Vikunja as Flutter cross platform app
 version: 0.1.8-beta
 
 environment:
-  sdk: ">=2.19.0 <3.0.0"
-  flutter: 3.22.2
+  sdk: ^3.9.0
```
kolaente, line 7: no more flutter dependency?

leschbenedikt: It still has a dependency on flutter.
This is only a version constraint which is not necessary and also only a lower bound

lib/core/background_work.dart
leschbenedikt: Done

lib/core/services.dart
leschbenedikt: Whole thing was unused. Think that was a merge issue on my side

lib/presentation/pages/login/register_page.dart
leschbenedikt: Good find. Came from the old version. Removed it.

lib/presentation/pages/project/project_edit.dart
leschbenedikt: Changed

leschbenedikt:
Is anything else missing?

kolaente:
Nothing missing, seems like I just forgot to merge this directly.

PR #150, opened 2025-09-30
Feature/error handling

leschbenedikt:


lib/core/network/client.dart
```diff
@@ -81,92 +81,94 @@ class Client {
...
-    headers.forEach((name, values) {
-      map[name] = values[0].toString();
-    });
-    return map;
-  }
+  Response<T> _handleResponse<T>(
+    http.Response response,
+    T Function(dynamic body)? mapper,
+  ) {
+    log.info("Processing response: ${response.body}");
```
kolaente: Should this be logged all the time or for dev only?

lib/core/network/client.dart
```diff
@@ -176,50 +178,35 @@ class Client {
...
-                      ),
-                    ),
-                  );
-                },
-              ),
-      );
-      if (showSnackBar) {
-        globalSnackbarKey.currentState?.showSnackBar(snackBar);
+      return ErrorResponse<T>(response.statusCode, headers, error);
+    } else {
```
kolaente: This is redundant since we return the line before

lib/core/network/client.dart
```diff
@@ -176,50 +178,35 @@ class Client {
...
+      var decode = utf8.decode(response.bodyBytes);
+      var convert = _decoder.convert(decode);
+      if (mapper != null) {
+        return SuccessResponse<T>(
+          mapper.call(convert),
+          response.statusCode,
+          response.headers,
+        );
       } else {
```
kolaente: same here

lib/core/utils/mapping_extensions.dart
```diff
@@ -0,0 +1,64 @@
...
+    if (isSuccessful) {
+      var successResponse = toSuccess();
+      var body = successResponse.body;
+
+      return SuccessResponse<T>(
+        body.toDomain(),
+        successResponse.statusCode,
+        successResponse.headers,
+      );
+    } else if (isError) {
```
kolaente, line 15: same here for early return.

This might be personal preference, coming from Go 😅

lib/presentation/manager/project_controller.dart
```diff
@@ -52,15 +89,16 @@ class ProjectController extends _$ProjectController {
 
     if (!displayDoneTasks) {
       queryParams.addAll({
-        "filter": ["done=false"],
+        "filter_by": ["done"],
+        "filter_value": ["false"],
```
kolaente: This should use the new filter syntax (as it was before)

kolaente:
When I try to edit a task I get this error:

leschbenedikt:
Fixed the edit page issue. Was for newly created task

lib/presentation/manager/project_controller.dart
leschbenedikt: Fixed. Found also that the filter was not working at also

kolaente:
Looks good to me, what's left to get this out of the draft state?

leschbenedikt:
I found another issue I will fix maybe two. Couldn't reproduce the second one until now. Could be an emulator related thing.
Also I want to test some more on my main device.

lib/core/network/client.dart
leschbenedikt: I chose info cause that's the default level that get upload in case of a error to entry.
The response could help a lot when analyzing errors

kolaente: The issue is this could contain really private data (task details), I don't think we should upload that to sentry.

leschbenedikt: You're right. I will remove it for now. Maybe we will find another solution for this

kolaente (approved):
Thanks!

PR #158, opened 2025-10-18
Several fixed & improvements

leschbenedikt:
- fixed the ios build, replaced downloader dep in that process
- added server version to sentry entries for better error analysis
- exluded several exception from sentry like wrong url input, timeouts etc.


ios/Flutter/Flutter.podspec
```diff
@@ -0,0 +1,18 @@
+#
+# This podspec is NOT to be published. It is only used as a local source!
+# This is a generated file; do not edit or check into version control.
```
kolaente: It says `do not check into version control` but it's fine to have here?

leschbenedikt: Not sure why this shouldn't be checked in as that's not critical info.
Also it not in the gitignore.
But removed it to be safe.

leschbenedikt:
Pushed some more fixes

kolaente (approved):
Thanks!

Issue #156, opened 2025-10-11
Adding a task in list view puts it at the bottom


kolaente:
When I create a task in list view, it is added at the bottom. When I then reload or reopen the project, the task is at the top (which is expected).

The task should be added to the top of the list, not at the bottom. The current behaviour is a little confusing and does not match what is happening in the web app.

Issue #116, opened 2025-04-12
Task description editor not functioning on pixel 6


rasmusravn:
I cannot edit task descriptions om my pixel 6 phone. Editor shows with options, but no text shows when there is a description already, and no keyboard pops up.

cool-nut:
The same issue. Unable to edit on the mobile device. No keyboard pops up and no cursor is displayed.

leschbenedikt:
I could reproduce it with the latest release but couldn't on the current Dev build. 
I think this is fixed, can someone confirm?

kolaente:
Follow up in https://github.com/go-vikunja/app/issues/175

Issue #154, opened 2025-10-11
Don't force users to select a due date when creating tasks


kolaente:
When creating a task, I have to select a due date. That's not always what I want, it should let me unselect the option.

![Screenshot_20251011-115531_1.png](https://github.com/user-attachments/assets/e8178267-7656-4e0c-8650-b5918d4200b2)

In the long run, we should bring quick add magic to the app, allowing to set a due date right when creating a task.

kolaente:
After closer inspection, it seems like the select does not do anything. I'd say we can just remove it.

leschbenedikt:
I think there was a bug at some time where it didn't work.
Should work know. I really like the feature and would keep it.

But I agree, it should be optional. I will think of a better solution.

kolaente:
Well the best way to solve this would be adding quick add magic, but that's a lot of complexity to integrate. Maybe just add an option in the settings for it?

leschbenedikt:
What do you think about that

<img width="327" height="287" alt="Image" src="https://github.com/user-attachments/assets/2bb91bd2-3090-43ea-8685-d9505daa3c26" />

<img width="327" height="345" alt="Image" src="https://github.com/user-attachments/assets/610ce746-0682-40a1-af55-6de39fe6ecd1" />

kolaente:
Looks great!

PR #162, opened 2025-10-22
Feature/UI improvements

### LINKED ISSUE #58: Project name in home screen
The home screen currently only displays the name and the due date of the tasks. This can be irritating if there are tasks with the same name belonging to different projects, e.g. two exercises named _Exercise 1_ belonging to two different university courses.

The web interface displays the corresponding project name of each task in the home / current tasks and upcoming screen.

### LINKED ISSUE #62: Reminders don't work
I've been looking for a self-hosted Todo app and Vikunja was the most recommended.
However since it was setup a month ago it's never worked to save reminders.
The Android app will let you go through the process but once saved it's not there on reopen.
The web for mobile doesn't even let you open tasks just change the due date.
You have to use the desktop web version which lets you set the reminder but there is never a notification.

I have the latest beta for Android notification fix.
On Pixel 7 Pro Android 14.

### LINKED ISSUE #113: Done Tasks still showing when "show done tasks" box is unchecked
The list view does not remove entries from view, when "show done tasks" box is unchecked. 
Opening the same project on desktop gives the desired view with only tasks that are not done. 

Thank you for an awesome program. 

Kind Regards
Rasmus

### LINKED ISSUE #127: App architecture
Hi,

as you may have noticed I recently started to contribute to this project.

But before I get really into this I wanted to know if there are any plans on app architecture?
At the moment I don't really see a clear pattern.

Currently it seems like most of the time the domain and data layers are mixed into the UI.
In my opinion the project could use some clear layers and structuring.

I would be happy to suggest a few things but wanted to check with you first before heading in the wrong direction.

### LINKED ISSUE #140: Improve error handling in case the server does not return json
Sentry Issue: [APP-58](https://vikunja.sentry.io/issues/6814657222/?referrer=github_integration)

```
TypeError: type 'Null' is not a subtype of type 'Map<String, dynamic>'
  File "user_data_source.dart", line 43, in UserDataSource.getCurrentUser.<fn>
  File "user_repository_impl.dart", line 26, in UserRepositoryImpl.getCurrentUser
  File "settings.dart", line 69, in SettingsPageState.init.<fn>
```

When the server does not return json (for example on proxy error), this error occurs. We should handle this and show a message to the user.

### LINKED ISSUE #148: Back button allows navigation past login screen, causing logout confusion
When pressing the back button repeatedly, it is possible to navigate past the login screen. This results in the user being logged out, which can cause confusion, especially if the user is on the "Home" screen and accidentally hits back too many times. The "Logout" button in the settings should be sufficient for logging out, and users should not be logged out via navigation actions.

(reported by a user via mail)

**Steps to reproduce**
1. Log in to the app and navigate to the "Home" screen.
2. Press the back button several times.

**Expected behavior**
- Pressing the back button should only navigate back to the "Home" view.
- Users should not be logged out by pressing back repeatedly; logout should only occur via the "Logout" button in the settings.

**Additional context**
This behavior can be confusing, as users might accidentally log out when trying to navigate back. The navigation stack should be limited so that the back button does not allow navigating past the "Home" view.

### LINKED ISSUE #155: "Tasks" headline should not show up when a project has only tasks and no sub projects
Feels a little redundant.

leschbenedikt:
- [x] related to #156 
- [x] Resolves #155 
- [x] Resolves #58
- [x] Closes #113
- [x] related #116
- [x] Closes #127 
- [x] Closes #140 
- [x] Closes #148 
- [x] Closes #62

### DIFF
--- a/lib/core/theming/theme.dart
+++ b/lib/core/theming/theme.dart
@@ -335,17 +335,16 @@ class MaterialTheme {
     return theme(darkHighContrastScheme());
   }
 
-
   ThemeData theme(ColorScheme colorScheme) => ThemeData(
-     useMaterial3: true,
-     brightness: colorScheme.brightness,
-     colorScheme: colorScheme,
-     textTheme: textTheme.apply(
-       bodyColor: colorScheme.onSurface,
-       displayColor: colorScheme.onSurface,
-     ),
-     scaffoldBackgroundColor: colorScheme.background,
-     canvasColor: colorScheme.surface,
+    useMaterial3: true,
+    brightness: colorScheme.brightness,
+    colorScheme: colorScheme,
+    textTheme: textTheme.apply(
+      bodyColor: colorScheme.onSurface,
+      displayColor: colorScheme.onSurface,
+    ),
+    scaffoldBackgroundColor: colorScheme.background,
+    canvasColor: colorScheme.surface,
   );
 
   /// success
--- a/lib/presentation/widgets/task/task_list_item.dart
+++ b/lib/presentation/widgets/task/task_list_item.dart
@@ -83,7 +83,7 @@ class TaskListItemState extends State<TaskListItem> {
     var project = task.project;
 
     if (texts.isEmpty) {
-      if(project != null) {
+      if (project != null) {
         return Text(project.title);
       }
 
@@ -95,11 +95,8 @@ class TaskListItemState extends State<TaskListItem> {
       child: Column(
         crossAxisAlignment: CrossAxisAlignment.start,
         children: [
-          if(project != null)
-            Text(
-              project.title,
-              style: Theme.of(context).textTheme.bodyMedium,
-            ),
+          if (project != null)
+            Text(project.title, style: Theme.of(context).textTheme.bodyMedium),
           Padding(
             padding: const EdgeInsets.only(top: 4.0),
             child: Row(children: texts),
--- a/lib/data/repositories/task_repository_impl.dart
+++ b/lib/data/repositories/task_repository_impl.dart
@@ -52,7 +52,10 @@ class TaskRepositoryImpl extends TaskRepository {
     String filterString, [
     Map<String, List<String>>? queryParameters,
   ]) async {
-    return (await _dataSource.getByFilterString(filterString, queryParameters)).toDomain();
+    return (await _dataSource.getByFilterString(
+      filterString,
+      queryParameters,
+    )).toDomain();
   }
 
   @override
--- a/lib/presentation/widgets/project/project_task_list_item.dart
+++ b/lib/presentation/widgets/project/project_task_list_item.dart
@@ -54,7 +54,11 @@ class ProjectTaskListItemState extends State<ProjectTaskListItem> {
             onPressed: () => widget.onEdit(),
           ),
         ),
-        Container(width: 4.0, height: subtitle != null ? 68.0 : 56.0, color: widget.task.color),
+        Container(
+          width: 4.0,
+          height: subtitle != null ? 68.0 : 56.0,
+          color: widget.task.color,
+        ),
       ],
     );
   }


lib/core/theming/theme.dart
```diff
@@ -0,0 +1,512 @@
...
+
+class MaterialTheme {
+  final TextTheme textTheme;
+
+  const MaterialTheme(this.textTheme);
+
+  static ColorScheme lightScheme() {
+    return const ColorScheme(
+      brightness: Brightness.light,
+      primary: Color(0xff0055c7),
```
kolaente: Is this used in similar places as the web ui? Because then it should be `#126cfd`

lib/presentation/widgets/task/task_list_item.dart
```diff
@@ -0,0 +1,104 @@
...
+          child: DueDateCard(task.dueDate!),
+        ),
+      );
+    }
+    if (task.priority != null && task.priority != 0) {
+      texts.add(PriorityBatch(task.priority!));
+    }
+
+    if (texts.isEmpty) {
+      return Text(task.project?.title ?? "-");
```
kolaente: Maybe we don't need to show a `-` when no project is available?

kolaente:
(not sure if this is done already or not since still draft)

- #156 is still a problem - the task is added at the bottom
- Same for #116 - no editor shown

I've converted the list of issues in the original PR description into a check list to make it easier to follow which are still open.

kolaente:
A few more things I noticed:

In dark mode, it looks like the due date has incorrect spacing - the pill shape of the background of the due date is not as visible as in light mode:



The task in the list view has superfluous spacing at the bottom:



Labels in kanban view need spacing:



On the home page, there is redundant spacing at the top:

lib/core/theming/theme.dart
leschbenedikt: The theme.dart is generated with https://material-foundation.github.io/material-theme-builder/

You put a seed in and a complete theme is generated.
It only adjusts minimally to match material specs.

I did the generation again which resulted in minor changes.

lib/presentation/widgets/task/task_list_item.dart
leschbenedikt: Done

leschbenedikt:
> * [Adding a task in list view puts it at the bottom #156](https://github.com/go-vikunja/app/issues/156) is still a problem - the task is added at the bottom

Fix another issue with the sorting when no overview project was set. Should be correct now.

leschbenedikt:
> * Same for [Task description editor not functioning on pixel 6 #116](https://github.com/go-vikunja/app/issues/116) - no editor shown

What is not working for. Are you unable to edit at all or is it just the cursor and keyboard part?

lib/presentation/widgets/task/task_list_item.dart
kolaente: It seems like now the project is only shown some times:



What are cases where we don't have the project? Is this a case where the API does not return it or do we need to load more data somehow?

kolaente:
> A few more things I noticed:

These seem fixed, great!

> Fix another issue with the sorting when no overview project was set. Should be correct now.

Unfortunately, still a problem

> What is not working for. Are you unable to edit at all or is it just the cursor and keyboard part?

I was able to get it working, but I found it highly unintuitive where to tap to get to the editor page. I'd say we should improve this, but might as well in a new PR.

leschbenedikt:
>I was able to get it working, but I found it highly unintuitive where to tap to get to the editor page. I'd say we should improve >this, but might as well in a new PR.

I agree, took me a while the first time also. I have some changes in mind. But as you said I would put the in another PR as this is no actual bug

leschbenedikt:
> > Fix another issue with the sorting when no overview project was set. Should be correct now.
> 
> Unfortunately, still a problem

That'S strange I can't reproduce it. For me the tasks are added according to the due date.
The sorting is
`"sort_by": ["due_date", "id"],
  "order_by": ["asc", "desc"],`
              
which seems to match the web version.

Does it still always put them at the bottom of the list for you?

kolaente:
> Does it still always put them at the bottom of the list for you?

It seems to depend on whether I add a due date or not. If I add a due date, the task is correctly sorted by due date, otherwise it appears at the bottom. I can't seem to _not_ explicitly add a due date, just when I'm in the list view and don't select anything (even though it appears selected by default), no due date is added. This seems a little broken. And then there's https://github.com/go-vikunja/app/issues/154 - IMHO, we should remove the due date selection when creating a task and then check if the sorting problem is still there. But that could be a separate PR as well.

For the list view in a project, it should sort the tasks by the position that was set manually as on the web, but we don't have manual sort yet.

Maybe let's discuss this elsewhere?

leschbenedikt:
Yes sound good. I will open a new issue for the sorting.
This PR is pretty big already

leschbenedikt:
Should we merge this now?

kolaente:
It seems like there are conflicts, please rebase and resolve them, then we're good to go

Issue #152, opened 2025-10-08
Quick add widget on homescreen


maltegartner:
Add a Quick Add widget button to home screen on Android. It means you don't have to enter the app to add something to the inbox, but can do it quickly and on the go.

In GTD, a core principle is to lower resistance to capturing to raise the likelihood of capturing everything that occupies your attention. I've found that Quick Add widget on my home screen makes me capture much more during the day.

kolemcrae:
Fully agree - thought it should likely be an Andoid shortcut - much more universal than a home page widget since you can easily assign the shortcut to anything.

maltegartner:
I'm a supporter for Vikunja, but not actually a user before features like these are built. I hope they will and would love to use a lightweight open source task manager.

brokoli18:
I am currently working on a widget for Vikunja, one of the capabilities will be a quick add button that will take you to the add-task page

maltegartner:
Awesome to hear @brokoli18 . If I may add: It would be a big improvement if the widget never had to take you into the actual program, but that you can just click the Quick add button on your homescreen, a small input screen pops up, you type in what you want to capture, press enter and the input screen disappears. You never experience having entered any program. The key to a good capture function in any task management program is that the level of resistance to capturing is as small as possible - as few clicks as possible, as fast as possible.

brokoli18:
@maltegartner I agree, it would be ideal if there was some kind of popup on the widget that made adding tasks even easier. That was actually my original intention (something like todoist has), but its a little more difficult than I thought. I want to do it eventually but its just much easier (for now) to have a button that opens the add-task screen within the app, which at least gives the user some way of creating a task (indirectly) from the widget.

PR #168, opened 2025-11-08
feat: add the ability to complete tasks directly from the widget. 

brokoli18:
Also fixed a circular dependency bug with background refresh.

Resolves https://github.com/go-vikunja/app/issues/152

### DIFF
--- a/lib/presentation/pages/task/task_list_page.dart
+++ b/lib/presentation/pages/task/task_list_page.dart
@@ -9,7 +9,6 @@ import 'package:vikunja_app/core/di/repository_provider.dart';
 import 'package:vikunja_app/domain/entities/task.dart';
 import 'package:vikunja_app/domain/entities/task_page_model.dart';
 import 'package:vikunja_app/presentation/manager/task_page_controller.dart';
-import 'package:vikunja_app/presentation/manager/widget_controller.dart';
 import 'package:vikunja_app/presentation/pages/error_widget.dart';
 import 'package:vikunja_app/presentation/pages/loading_widget.dart';
 import 'package:vikunja_app/presentation/pages/task/task_edit_page.dart';


android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -37,36 +47,87 @@ class AppWidget : GlanceAppWidget() {
...
+
 
+    private fun doneTask(context: Context, prefs: SharedPreferences, taskID: String) {
+        prefs.edit {
+            putString("completeTask", taskID)
+            commit()
+        }
+        val backgroundIntent = HomeWidgetBackgroundIntent.getBroadcast(
+            context,
+            Uri.parse("appWidget://completeTask")
```
kolaente: Is this a general thing or should this be a custom protocol for the Vikunja app? If the latter, shouldn't this be something like `vikunja-app://completeTask`?

(I don't know much about Android, it's entirely valid if the way you did it is the correct way)

android/app/src/main/AndroidManifest.xml
```diff
@@ -23,14 +23,21 @@
         android:icon="@mipmap/ic_launcher"
         android:networkSecurityConfig="@xml/network_security_config"
         >
-    <receiver android:name=".AppWidgetReciever"
-          android:exported="true">
+        <receiver android:name="es.antonborri.home_widget.HomeWidgetBackgroundReceiver"  android:exported="true">
             <intent-filter>
-            <action android:name="android.appwidget.action.APPWIDGET_UPDATE" />
+                <action android:name="es.antonborri.home_widget.action.BACKGROUND" />
```
kolaente, line 28: Where is `es.antonborri.home_widget` coming from? Shouldn't this be app-specific as well?

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
brokoli18: This is a general thing. i will change this to vikunja-app as requested

android/app/src/main/AndroidManifest.xml
brokoli18: This is coming from the the home_widget package (https://github.com/ABausG/home_widget) which abstracts away some of the complexity necessary to communicate from kotlin to flutter. According to the documentation it has to be exactly like that

lib/presentation/pages/task/task_list_page.dart
```diff
@@ -47,6 +48,7 @@ class TaskListPageState extends ConsumerState<TaskListPage> {
 
     return pageModel.when(
       data: (model) {
+        updateWidget();
```
leschbenedikt: That would be better suited int the task controller.

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
leschbenedikt: Can we use a parameter for the task in Uri? This could result in errors if I complete multiple task very fast

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -37,36 +47,87 @@ class AppWidget : GlanceAppWidget() {
...
             }
+
         }
     }
 
     @Composable
     private fun WidgetTitleBar() {
         Box(
-            modifier = GlanceModifier.fillMaxWidth().height(50.dp).background(Color(0xFF0c86ff)),
+            modifier = GlanceModifier.fillMaxWidth().height(50.dp).background(Color.Blue),
```
leschbenedikt: Hey,
could you please use the colors of the theme and support dark mode.
Something like this `.background(ColorProvider(Color(0xFF126cfd), Color(0xFF013992)))`

Will also be necessary for the list background.

Thanks

lib/presentation/pages/task/task_list_page.dart
brokoli18: @leschbenedikt I have moved this to the task controller. Have a look and tell me if you are happy with its location now

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
brokoli18: @leschbenedikt I changed the Color to one of the hexes you provided. As far as I know the background (at least for the bright theme) is just white? Let me know if that is not the case.

As regards for supporting dark mode I think this is outside the scope of this MR. This is just for completing tasks, we can add theme support in a separate change.

brokoli18:
@kolaente Just FYI this MR will not resolve https://github.com/go-vikunja/app/issues/152 as this will just let a user complete tasks (not add them). I am working on a separate change that will also let a user add tasks from the widgets (indirectly). Its almost ready to go

lib/presentation/pages/task/task_list_page.dart
leschbenedikt: Unfortunately not 😄 
Get methods should only "retrieve" something not actively change something.
Otherwise it can be really hard to understand the code and/or debug.

I'm not saying the I do it all the time but when noticed it should be changed 

I would suggest moving it up one call

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
leschbenedikt: I'm fine with that. We can create another issue for adding darkmode to the widget

lib/presentation/pages/task/task_list_page.dart
brokoli18: Do you mean moving it into the `reload()` and `build()` methods?

leschbenedikt: Yes that would be the "right" spot, cause that are the methods that "manage" what to do when we want updated task.

It think there is more potential for refactors but that's something for another day.

brokoli18: @leschbenedikt Ive moved it up and tested to check that the widget is still updated appropriately.

brokoli18:
@leschbenedikt @kolaente The above check is failing because now dart is saying that it wants to format 157 files? I dont think we want to touch so many files on this mr.

kolaente:
@brokoli18 it said only one was changed: `lib/presentation/manager/task_page_controller.dart`.

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/Task.kt
```diff
@@ -0,0 +1,12 @@
...
+import androidx.annotation.Keep
+import java.util.Date
+
+@Keep
+class Task (
+    var id: String,
+    var title: String,
+    var dueDate: Date,
+    var today: Boolean,
+    var overdue: Boolean,
```
leschbenedikt: This seems to be unused

lib/presentation/manager/widget_controller.dart
```diff
@@ -3,16 +3,64 @@ import 'dart:developer' as developer;
...
 import 'package:vikunja_app/domain/entities/task.dart';
+import 'package:vikunja_app/domain/entities/widget_task.dart';
 import 'package:vikunja_app/domain/repositories/task_repository.dart';
 
-List<Task> filterForTodayTasks(List<Task> tasks) {
+void completeTask(String taskID) async {
+  if (taskID == "null") {
+    developer.log("Tried to complete an empty task");
+  }
+  ;
```
leschbenedikt: Unnecessary semicolon

lib/presentation/manager/widget_controller.dart
```diff
@@ -3,16 +3,64 @@ import 'dart:developer' as developer;
...
+
+WidgetTask convertTask(Task task) {
+  // Check if task is for today
+  final now = DateTime.now();
+  final today = DateTime(now.year, now.month, now.day);
+
+  bool wgToday = task.dueDate!.day == today.day ? true : false;
+  bool overdue = task.dueDate!.isBefore(now) ? true : false;
+
+  // CHeck if task is overdue
```
leschbenedikt: The placing of the comment seems to be wrong

lib/presentation/manager/widget_controller.dart
```diff
@@ -52,22 +100,25 @@ Future<void> updateWidget() async {
...
+  HomeWidget.saveWidgetData("WidgetTaskIDs", widgetTaskIDs.toString());
+  reRenderWidget();
+}
 
-  // Update the widget
+void reRenderWidget() {
   HomeWidget.updateWidget(
     name: 'AppWidget',
-    qualifiedAndroidName: 'io.vikunja.flutteringvikunja.AppWidgetReciever',
+    // androidName: '.widget.AppWidgetReciever',
```
leschbenedikt: Please remove this line if not necessary

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -37,36 +48,88 @@ class AppWidget : GlanceAppWidget() {
         }
     }
 
+    // This function cannot be composable otherwise it wont run sometimes when shared prefs isnt changed
```
leschbenedikt: I don't understand this comment.
Composable are just widgets. So why should this function be composable at all?

lib/presentation/manager/widget_controller.dart
```diff
@@ -3,16 +3,64 @@ import 'dart:developer' as developer;
...
+    Client client = Client(token: token, base: base);
+    tz.initializeTimeZones();
+
+    var ignoreCertificates = await datasource.getIgnoreCertificates();
+    client.setIgnoreCerts(ignoreCertificates);
+
+    TaskRepository taskService = TaskRepositoryImpl(TaskDataSource(client));
+    var taskResponse = await taskService.getTask(int.parse(taskID!));
+    var task = taskResponse.toSuccess().body;
+    taskService.update(task.copyWith(done: true));
```
leschbenedikt: You should wait for this to finish. Otherwise the update of the widget could occur before the task is actually finished

lib/presentation/manager/settings_controller.dart
```diff
@@ -110,36 +110,36 @@ class SettingsController extends _$SettingsController {
...
-            "update-tasks",
-            "update-tasks",
-            frequency: duration,
-            constraints: Constraints(networkType: NetworkType.connected),
-            initialDelay: Duration(seconds: 15),
-          );
-        }
-
+    var settings = await getAll();
+    // var settings = ref.read(settingsControllerProvider);
```
leschbenedikt: Please remove this lines if not necessary

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -37,36 +48,88 @@ class AppWidget : GlanceAppWidget() {
...
+    @Composable
+    private fun GlanceContent(context: Context, currentState: HomeWidgetGlanceState) {
+        val prefs = currentState.preferences
+        getTasks(prefs)
         Column {
             WidgetTitleBar()
-            LazyColumn(modifier = GlanceModifier.background(Color.White)) {
-                items(tasks) { task ->
-                    RenderRow(task)
+            if (todayTasks.isNotEmpty() or otherTasks.isNotEmpty()) {
```
leschbenedikt: If today task are empty the header is still visible without items.
The same happens when overdue is empty

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -37,36 +48,88 @@ class AppWidget : GlanceAppWidget() {
...
+                LazyColumn(modifier = GlanceModifier.background(Color.White)) {
+                    item{
+                        Text("Today:")
+                    }
+                    items(todayTasks.sortedBy { it.dueDate }) { task ->
+                        RenderRow(context, task, prefs, "HH:mm")
+                    }
+                    item{
+                        Text("OverDue")
```
leschbenedikt: Misspelled

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -37,36 +48,88 @@ class AppWidget : GlanceAppWidget() {
...
+        val taskURI = uri.buildUpon().appendQueryParameter("taskID", taskID).build()
+        val backgroundIntent = HomeWidgetBackgroundIntent.getBroadcast(
+            context,
+            taskURI
+        )
+        backgroundIntent.send()
+    }
 
+    @Composable
+    private fun GlanceContent(context: Context, currentState: HomeWidgetGlanceState) {
```
leschbenedikt, line 101: The size and position of the widget changes depending of the number of task.
I think this should always be the same.

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -37,36 +48,88 @@ class AppWidget : GlanceAppWidget() {
...
-            LazyColumn(modifier = GlanceModifier.background(Color.White)) {
-                items(tasks) { task ->
-                    RenderRow(task)
+            if (todayTasks.isNotEmpty() or otherTasks.isNotEmpty()) {
+                LazyColumn(modifier = GlanceModifier.background(Color.White)) {
+                    item{
+                        Text("Today:")
+                    }
+                    items(todayTasks.sortedBy { it.dueDate }) { task ->
+                        RenderRow(context, task, prefs, "HH:mm")
```
leschbenedikt: The pattern shouldn't be hardcoded, the should depend on the locale.
You can just pick some of the predefined ones from flutter

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -37,36 +48,88 @@ class AppWidget : GlanceAppWidget() {
...
+                    }
+                    item{
+                        Text("OverDue")
+                    }
+                    items(otherTasks.sortedBy { it.dueDate }) { task ->
+                        RenderRow(context, task, prefs, "dd MMM HH:mm")
```
leschbenedikt: Same here

lib/core/background_work.dart
```diff
@@ -14,6 +14,14 @@ import 'package:vikunja_app/presentation/manager/notifications.dart';
 import 'package:vikunja_app/presentation/manager/widget_controller.dart';
 import 'package:workmanager/workmanager.dart';
 
+@pragma("vm:entry-point")
+Future<void> widgetCallback(Uri? uri) async {
+  if (uri?.host == "completetask") {
+    String? taskID = uri?.queryParameters['taskID'];
+    completeTask(taskID!);
```
leschbenedikt: If taskID is null just do nothing
Otherwise it will crash

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/Task.kt
brokoli18: This class is used when the widget pulls tasks from shared storage. I don't need most of the values from the default Task class for the widget so it was easier to create a new one

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
brokoli18: I will remove this comment as its not really necessary, it was mainly for me.

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
brokoli18: Can you elaborate on what you mean here. When the widget is first created its a 4x1 Rectangle. After the user resizes it it should stay the same size regardless of how many tasks there are.

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
brokoli18: I have looked at the possible solutions here, but using something like `DateTimeFormatter.ISO_LOCAL_DATE_TIME` the date is either too big for the widget, or doesn't display the right components. In short the string fomatter is just much more flexible, is there an issue with me using it here?

brokoli18:
@leschbenedikt I made some changes/comments, can you have a look when you have a moment

brokoli18:
@kolaente Is it possible to merge this MR? I have a couple of other features I want to add to the widget (Adding tasks from the widgets for example), which this MR is currently blocking

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
leschbenedikt: The height of the widget changed depending on the number of task displayed. And because it was centered the position also.

So if it displayed 2 tasks and you checked 1 it would shrink.
I changed that by always filling the viewport.

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
leschbenedikt: The problem with hardcoding is that the date is written different depending where you live.
For hour and minute it doesn't matter as far as I know.  But the most obvious thing is that some countries write the month first and some the day. 

Also there could differences we don't know about that could ultimately result in the date displayed "wrong".

I updated the code so it will display locale dependent

Issue #167, opened 2025-11-08
[Request] Remove Hard dependency on Google Play Services


RXTX4816:
**As is:**
----
The Vikunja App requires Google Play Store Services to be installed and running in the background in order to use the App.

**To be:**
----
The Vikunja App is usable without Google Play Store Services

**Background:**
----
[102](https://github.com/go-vikunja/app/issues/102)
Since this is an open-source application with zero direct usages of the Google Play Services API, and cronethttp is the only dependency that, without the build parameter flag `--dart-define=cronetHttpNoPlay=true`, requires GPlay Services.

Although the file size of the Application increases by 10 mb, I think we all have these 10 megabytes free on our current devices..

Thanks for listening

leschbenedikt:
I just tried to reproduce this and couldn't.
I used an emulator with the AOSP image which shouldn't have play services and it worked.
But it's just an emulator so I'm not sure.

Could you please try the newest version of the app and check if it works for you now.

RXTX4816:
I apologize, I tested it and you are right, it works. 

I just tested it on Waydroid LineageOS Emulator without GPlay and it works and connects. I don't know how to read App logs in Waydroid.

So I tried it now on a real stock AOSP device that has no gplay and it also connects successfully and everything seems to work.
On this device I was able to extract the app logs, both for the NoGplay(my PR) version and the default version.

It actually makes no difference whether the flag is issued, both versions produce Warnings and Errors related to GPlay and GStore when no GPlay is installed, so there might be further libraries that depend on GooglePlayServices in the background

 ```
--------- switch to main
11-10 11:55:43.898 32738 32738 I ImeTracker: system_server:170023d0: onCancelled at PHASE_CLIENT_ON_CONTROLS_CHANGED
11-10 11:55:44.056 32738 32738 W GooglePlayServicesUtil: io.vikunja.app requires the Google Play Store, but it is missing.
11-10 11:55:44.056 32738 32738 E GooglePlayServicesUtil: GooglePlayServices not available due to error 9
11-10 11:55:45.479 32738 32738 W GooglePlayServicesUtil: io.vikunja.app requires the Google Play Store, but it is missing.
11-10 11:55:45.479 32738 32738 E GooglePlayServicesUtil: GooglePlayServices not available due to error 9
11-10 11:55:45.479 32738 32738 I PlayServicesCronet: Google-Play-Services-Cronet-Provider is unavailable.

```

leschbenedikt:
[...]

PR #171, opened 2025-11-09
fix: do not rebuild client on every request

RXTX4816:
https://github.com/go-vikunja/app/issues/167


leschbenedikt:
Although this works I wouldn't recommend this solution.
Riverpod does have build in mechanics for that.

I would be a better solution to use them.

RXTX4816:
Added riverpod httpclient handling

leschbenedikt:
I thought more of something like this.
It's easier and doesn't need the di to know the details of the implementation

RXTX4816:
Thats a clean solution, thanks!

leschbenedikt (approved):

PR #174, opened 2025-11-10
feat: allow user to set no due date when creating a task

### LINKED ISSUE #154: Don't force users to select a due date when creating tasks

leschbenedikt:
Resolves #154

### DIFF
--- a/lib/domain/entities/new_task_due.dart
+++ b/lib/domain/entities/new_task_due.dart
@@ -40,7 +40,8 @@ enum NewTaskDue {
       case NewTaskDue.later_this_week:
         return newDateTime.add(
           Duration(
-            days: currentDateTime.weekday == DateTime.friday ||
+            days:
+                currentDateTime.weekday == DateTime.friday ||
                     currentDateTime.weekday == DateTime.saturday ||
                     currentDateTime.weekday == DateTime.sunday
                 ? 0
--- a/test/new_task_due_test.dart
+++ b/test/new_task_due_test.dart
@@ -34,7 +34,13 @@ void main() {
     });
 
     test('next_monday when today is Wednesday returns next Monday', () {
-      final wednesday = DateTime(2025, 11, 12, 13, 0); // Wednesday 13:00 -> nearest 15
+      final wednesday = DateTime(
+        2025,
+        11,
+        12,
+        13,
+        0,
+      ); // Wednesday 13:00 -> nearest 15
       final result = NewTaskDue.next_monday.calculateDate(wednesday);
       expect(result!.day, wednesday.day + 5); // Wednesday -> next Monday
       expect(result.weekday, DateTime.monday);
@@ -42,7 +48,13 @@ void main() {
     });
 
     test('weekend when today is Monday returns Saturday', () {
-      final monday = DateTime(2025, 11, 10, 11, 0); // Monday 11:00 -> nearest 12
+      final monday = DateTime(
+        2025,
+        11,
+        10,
+        11,
+        0,
+      ); // Monday 11:00 -> nearest 12
       final result = NewTaskDue.weekend.calculateDate(monday);
       expect(result!.weekday, DateTime.saturday);
       expect(result.day, monday.day + 5);
@@ -50,14 +62,26 @@ void main() {
     });
 
     test('weekend when today is Saturday returns today', () {
-      final saturday = DateTime(2025, 11, 15, 20, 0); // Saturday 20:00 -> nearest 21
+      final saturday = DateTime(
+        2025,
+        11,
+        15,
+        20,
+        0,
+      ); // Saturday 20:00 -> nearest 21
       final result = NewTaskDue.weekend.calculateDate(saturday);
       expect(result!.day, saturday.day);
       expect(result.hour, 21);
     });
 
     test('later_this_week early in week adds two days', () {
-      final tuesday = DateTime(2025, 11, 11, 14, 59); // Tuesday 14:59 -> nearest 15
+      final tuesday = DateTime(
+        2025,
+        11,
+        11,
+        14,
+        59,
+      ); // Tuesday 14:59 -> nearest 15
       final result = NewTaskDue.later_this_week.calculateDate(tuesday);
       expect(result!.day, tuesday.day + 2);
       expect(result.hour, 15);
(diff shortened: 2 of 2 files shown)


lib/domain/entities/new_task_due.dart
```diff
@@ -1,19 +1,69 @@
...
+            .copyWith(hour: hour, minute: 0, second: 0)
+            .add(Duration(days: (DateTime.saturday - dateTime.weekday) % 6));
+      case NewTaskDue.later_this_week:
+        return dateTime
+            .copyWith(hour: hour, minute: 0, second: 0)
+            .add(
+              Duration(
+                days:
+                    dateTime.day == DateTime.friday ||
+                        dateTime.day == DateTime.saturday ||
```
RXTX4816: Bug: should be dateTime.weekday

lib/domain/entities/new_task_due.dart
```diff
@@ -1,19 +1,69 @@
...
+            .copyWith(hour: hour, minute: 0, second: 0)
+            .add(Duration(days: 1));
+      case NewTaskDue.next_monday:
+        return dateTime
+            .copyWith(hour: hour, minute: 0, second: 0)
+            .add(Duration(days: (DateTime.monday - dateTime.weekday) % 7));
+      case NewTaskDue.weekend:
+        return dateTime
+            .copyWith(hour: hour, minute: 0, second: 0)
+            .add(Duration(days: (DateTime.saturday - dateTime.weekday) % 6));
```
RXTX4816: `DateTime.saturday - dateTime.weekday) % 6`

Lets say its sunday=7 now:

6 - 7 = -1 
-1 % 6 = 5

But Saturday is 6 days ahead of Sunday, not 5 :)

lib/domain/entities/new_task_due.dart
```diff
@@ -1,19 +1,69 @@
...
+                days:
+                    dateTime.day == DateTime.friday ||
+                        dateTime.day == DateTime.saturday ||
+                        dateTime.day == DateTime.sunday
+                    ? 0
+                    : 2,
+              ),
+            );
+      case NewTaskDue.next_week:
+        return dateTime.copyWith(hour: hour).add(Duration(days: 7));
```
RXTX4816: This does not include minute/seconds as the other ones do, was this intentional?

lib/presentation/widgets/task/add_task_dialog.dart
```diff
@@ -91,22 +121,21 @@ class AddTaskDialogState extends State<AddTaskDialog> {
...
+    return ChoiceChip(
+      label: Text(name),
+      selected: newTaskDue == thisNewTaskDue,
+      onSelected: (value) {
+        newTaskDue = thisNewTaskDue;
+        setState(() {
+          if (newTaskDue == NewTaskDue.custom ||
+              newTaskDue == NewTaskDue.none) {
+            dueDate = null;
+            dueDate = null;
```
RXTX4816: Can remove one duplicate

RXTX4816 (changes requested):
Great work! I like this feature a lot. 

I reviewed your code and noted some stuff that should probably be changed.

kolaente:
Tested, but it seems like the due date is only set when creating a task one the home page. No due date is ever added when inside a project.

lib/domain/entities/new_task_due.dart
leschbenedikt: Fixed

lib/domain/entities/new_task_due.dart
leschbenedikt: Fixed

lib/domain/entities/new_task_due.dart
leschbenedikt: No it was not, just missed a case. Fixed

lib/presentation/widgets/task/add_task_dialog.dart
leschbenedikt: Fixed

leschbenedikt:
> Tested, but it seems like the due date is only set when creating a task one the home page. No due date is ever added when inside a project.

You're right. Never spotted this sind I always use the main page.
Is fixed now

lib/domain/entities/new_task_due.dart
RXTX4816: Approved ✅

lib/domain/entities/new_task_due.dart
RXTX4816: ✅

lib/domain/entities/new_task_due.dart
RXTX4816: Nice! 
There is a small discrepancy with the web currently;

Let it be 14:55 now.

Web: Calculates it as 15:00 (Since its closer to 15:00 than 14:00)
App: Calculates it as 14:00 (Copy hour, strip minute and second)

leschbenedikt: I don't know what you mean. 14 is no possible value. Only 9, 12, 15, 18 and 21

lib/domain/entities/new_task_due.dart
```diff
@@ -1,19 +1,76 @@
...
+  today,
+  tomorrow,
+  next_monday,
+  weekend,
+  later_this_week,
+  next_week,
   custom;
 
-  Duration newTaskDueToDuration() {
+  DateTime? calculateDate() {
```
kolaente: Can you add unit tests for this? (perfect use case to ask an AI coding agent)

kolaente:
Looks good otherwise!

lib/domain/entities/new_task_due.dart
RXTX4816: You are right there is no issue here looks good 👍

lib/domain/entities/new_task_due.dart
leschbenedikt: Yeah sure, will do so

leschbenedikt: Added test cases with the help of an AI agent.
 Needed to adjust a few things but in general pretty good

PR #176, opened 2025-11-10
fix: sharing text with app

### LINKED ISSUE #157: Sharing to Vikunja causes error
When I try to share a page or something to Vikunja it throws an error. The error goes by so fast I can not read it. 

Expected: url of website now available as task that I can save in the inbox. 

Actual: It just throws an error and leaves me at the inbox.

leschbenedikt:
Resolves #157

### DIFF
--- a/lib/presentation/pages/home_page.dart
+++ b/lib/presentation/pages/home_page.dart
@@ -1,3 +1,5 @@
+import 'dart:developer' as developer;
+
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
@@ -103,14 +105,12 @@ class HomePageState extends ConsumerState<HomePage> {
     try {
       String? argument = await platform.invokeMethod<String>("isQuickTile", "");
 
-      print("Method $argument");
       return showAddItemDialog(argument);
     } catch (e) {
-      print("Error $e");
+      developer.log("Error $e");
     }
 
     platform.setMethodCallHandler((call) async {
-      print("Called $call");
       return showAddItemDialog(call.arguments as String);
     });
   }


android/app/src/main/kotlin/io/vikunja/flutteringvikunja/MainActivity.kt
```diff
@@ -1,54 +1,100 @@
...
         val action: String? = intent.action
         val type: String? = intent.type
-        val channel = MethodChannel(flutterEngine.dartExecutor.binaryMessenger, CHANNEL)
 
-        if ("ACTION_INSERT" == action && type != null && "ADD_NEW_TASK" == type) {
-            if (isNewIntent) channel.invokeMethod("open_add_task", "")
-            launchMethod = "open_add_task";
+        when (action) {
+            Intent.ACTION_INSERT -> {
+                if ("ADD_NEW_TASK" == type) {
```
kolaente: Maybe worth adding a const for these types?

lib/presentation/pages/home_page.dart
```diff
@@ -88,6 +99,91 @@ class HomePageState extends ConsumerState<HomePage> {
     return widgets[pos];
   }
 
+  void scheduleIntent() async {
+    try {
+      String? argument = await platform.invokeMethod<String>("isQuickTile", "");
+
+      print("Method $argument");
```
kolaente: shouldn't this use proper logging?

kolaente (approved):
Thanks

kolaente:
I noticed that when sharing multiple lines of text, only the first line would be shown in the add dialogue but all lines will be added as the task title. I assume that's a different problem, will open an issue.

EDIT: https://github.com/go-vikunja/app/issues/178

PR #177, opened 2025-11-11
feat: i18n language support

### LINKED ISSUE #59: i18n?
It would be great to add multi-language support and make it easier for community to translate.

RXTX4816:
Resolves #59 

Here is one version of how this app would look like with language support. 
Minimal testing shows its working fine.
Please review code/adapt changes.

- One .arb file per language (Currently: en, pl)
- `flutter gen-l10n` must be run to generate dart files from .arb files. (included in cicd and makefile) 
- Chooses System Default language, can be overriden in Vikunja Settings (Dropdown)
- If the System Default Language is not supported, it defaults to english.
- Minimal Test Cases to test if each language that has .arb files can be loaded.

Note: Somehow github does not show the file differences correctly, but VSCode does.

I think something like that would be a great addition to the first release and enable others to add their languages easily.

### DIFF
--- a/lib/core/utils/language_autonyms.dart
+++ b/lib/core/utils/language_autonyms.dart
@@ -3,63 +3,7 @@ import 'package:intl/intl.dart';
 
 const Map<String, String> kLocaleAutonyms = <String, String>{
   'en': 'English',
-  'en_US': 'English (United States)',
-  'en_GB': 'English (United Kingdom)',
-  'es': 'español',
-  'es_ES': 'español (España)',
-  'pl': 'polski',
-  'pl_PL': 'polski (Polska)',
-  'de': 'Deutsch',
-  'de_DE': 'Deutsch (Deutschland)',
-  'fr': 'français',
-  'fr_FR': 'français (France)',
-  'pt': 'português',
-  'pt_PT': 'português (Portugal)',
-  'pt_BR': 'português (Brasil)',
-  'it': 'italiano',
-  'it_IT': 'italiano (Italia)',
-  'nl': 'Nederlands',
-  'nl_NL': 'Nederlands (Nederland)',
-  'ru': 'русский',
-  'ru_RU': 'русский (Россия)',
-  'uk': 'українська',
-  'uk_UA': 'українська (Україна)',
-  'tr': 'Türkçe',
-  'tr_TR': 'Türkçe (Türkiye)',
-  'el': 'Ελληνικά',
-  'el_GR': 'Ελληνικά (Ελλάδα)',
-  'cs': 'čeština',
-  'cs_CZ': 'čeština (Česká republika)',
-  'sk': 'slovenčina',
-  'sk_SK': 'slovenčina (Slovensko)',
-  'sv': 'svenska',
-  'sv_SE': 'svenska (Sverige)',
-  'da': 'dansk',
-  'da_DK': 'dansk (Danmark)',
-  'nb': 'norsk bokmål',
-  'nb_NO': 'norsk bokmål (Norge)',
-  'fi': 'suomi',
-  'fi_FI': 'suomi (Suomi)',
-  'hu': 'magyar',
-  'hu_HU': 'magyar (Magyarország)',
-  'ro': 'română',
-  'ro_RO': 'română (România)',
-  'bg': 'български',
-  'bg_BG': 'български (България)',
-  'sr_Latn': 'srpski (latinica)',
-  'sr_Cyrl': 'српски (ћирилица)',
-  'hr': 'hrvatski',
-  'hr_HR': 'hrvatski (Hrvatska)',
-  'bs': 'bosanski',
-  'bs_BA': 'bosanski (Bosna i Hercegovina)',
-  'sl': 'slovenščina',
-  'sl_SI': 'slovenščina (Slovenija)',
-  'lt': 'lietuvių',
-  'lt_LT': 'lietuvių (Lietuva)',
-  'lv': 'latviešu',
-  'lv_LV': 'latviešu (Latvija)',
-  'et': 'eesti',
-  'et_EE': 'eesti (Eesti)',
+  'pl': 'Polski',
 };
 
 String languageAutonym(Locale locale) {
@@ -71,7 +15,7 @@ String languageAutonym(Locale locale) {
       nativeNames[locale.languageCode.toLowerCase()];
 
   if (autonym != null && autonym.isNotEmpty) {
-    return autonym[0].toUpperCase() + autonym.substring(1);
+    return autonym;
   }
 
   final country = locale.countryCode;
--- a/lib/l10n/app_es.arb
+++ b/lib/l10n/app_es.arb
@@ -1,4 +1,0 @@
-{
-  "@@locale": "es",
-  "welcomeMessage": "Bienvenido a Vikunja"
-}
--- a/lib/presentation/pages/home_page.dart
+++ b/lib/presentation/pages/home_page.dart
@@ -195,9 +195,9 @@ class HomePageState extends ConsumerState<HomePage> {
   }
 
   Future<void> postVersionCheckSnackbar() async {
-    var latestVersionTag = await ref
-        .read(versionRepositoryProvider)
-        .getLatestVersionTag();
+    var latestVersionTag =
+        await ref.read(versionRepositoryProvider).getLatestVersionTag() ??
+        "N/A";
     ref.read(versionRepositoryProvider).isUpToDate().then((value) {
       if (!value) {
         // not up to date


RXTX4816:
IMG_20251112_004054
IMG_20251112_003913

leschbenedikt:
Awesome work. I tried so review this as soon as possible but it will take some time.

lib/core/utils/priority.dart
```diff
@@ -1,37 +1,69 @@
...
-    case 'DO NOW':
-      return 5;
-    default:
-      // unset
-      return 0;
-  }
-}
+import 'package:vikunja_app/l10n/gen/app_localizations.dart';
+
+priorityToString(int? priority) {
```
leschbenedikt: I don't think we need both the localized and non localized version.

I would suggest replacing the old ones with the new ones, so the tests will be executed again.
Also I don't think the naming should contain the L10n part.
In a localized app I would assume the every string is returned localized if needed

lib/data/data_sources/settings_data_source.dart
```diff
@@ -147,4 +147,17 @@ class SettingsDatasource {
   Future<void> saveUserToken(String? token) {
     return _storage.write(key: "user-token", value: token);
   }
+
+  // Locale override (if null or empty -> use system locale)
+  Future<String?> getLocaleOverride() async {
+    return _storage.read(key: "locale_override");
+  }
+
+  Future<void> setLocaleOverride(String? localeCode) async {
```
leschbenedikt, line 156: You don't need this check. You can safely write null values.
If you check the source of the write method you will see it does the exact same think you did :)

lib/main.dart
```diff
@@ -1,95 +1,108 @@
...
+
+  var notifDenies = await Permission.notification.isDenied;
+  if (notifDenies) {
+    Permission.notification.request();
+  }
+
+  try {
+    if (!kIsWeb) {
+      FileDownloader().configureNotification(
+        running: TaskNotification('Downloading', 'file: {filename}'),
```
leschbenedikt: Missing translation

lib/presentation/manager/task_page_controller.dart
```diff
@@ -4,7 +4,7 @@ import 'package:vikunja_app/core/di/repository_provider.dart';
 import 'package:vikunja_app/core/network/response.dart';
 import 'package:vikunja_app/domain/entities/task.dart';
 import 'package:vikunja_app/domain/entities/task_page_model.dart';
-import 'package:vikunja_app/presentation/manager/projects_controller.dart';
+// import 'package:vikunja_app/presentation/manager/projects_controller.dart';
```
leschbenedikt: Just remove it when it's unused

lib/presentation/pages/error_widget.dart
```diff
@@ -1,54 +1,55 @@
...
+          ],
+        ),
+      ),
+    );
+  }
+
+  Widget getErrorWidget(BuildContext context, Object error) {
+    if (error is ClientException) {
+      return Text(
+        "Connection error",
```
leschbenedikt: Missing translation

kolaente:
Nice work! We should put the translations on crowdin and sync them here, like we do with the web UI but I can do that in a follow-up PR once this is merged.

lib/core/di/locale_provider.dart
```diff
@@ -0,0 +1,30 @@
...
+import 'package:vikunja_app/domain/repositories/settings_repository.dart';
+
+/// Provides an optional locale override. When null, system locale is used.
+final localeOverrideProvider =
+    StateNotifierProvider<LocaleOverrideNotifier, Locale?>((ref) {
+      final settingsRepo = ref.read(settingsRepositoryProvider);
+      return LocaleOverrideNotifier(settingsRepo);
+    });
+
+class LocaleOverrideNotifier extends StateNotifier<Locale?> {
```
leschbenedikt: StateNotifier is outdated. Also we do use riverpod_generator.

How a look here on how to migrate:
https://riverpod.dev/docs/migration/from_state_notifier

lib/presentation/pages/settings_page.dart
```diff
@@ -1,270 +1,307 @@
...
+                    .setDefaultProject(value);
+              }
+            },
+          ),
+        ),
+      ],
+    );
+  }
+
+  String _localeName(AppLocalizations l10n, Locale locale) {
```
leschbenedikt: Not sure about this but should we translate languages?
Could be hard to switch back if you misclicked.
But also weird to have every language in native.

What do you think about this?

kolaente: We have the languages in their native language in the web, I would argue let's do it in the app as well so that it's consistent

RXTX4816: Interesting question. 
In my opinion native languages are fine, but we should leave the first entry "System Langauge (default)" and not translate only this one and maybe name it like "System default".
This way, if someone accidently selects a non-roman alphabet langauge he can revert the change without a headache.

leschbenedikt: I agree with kolaente that we probably do it the same way as in the web version.
I we have good reason to do it different we should do it on both versions

lib/core/utils/priority.dart
RXTX4816: ✅

lib/data/data_sources/settings_data_source.dart
RXTX4816: ✅

lib/main.dart
RXTX4816: ✅
also a missing one in another file

lib/presentation/manager/task_page_controller.dart
RXTX4816: ✅

lib/presentation/pages/error_widget.dart
RXTX4816: ✅

lib/core/di/locale_provider.dart
RXTX4816: ✅

lib/presentation/pages/settings_page.dart
RXTX4816: ✅

.github/workflows/test.yml
```diff
@@ -1,70 +1,74 @@
...
+      - name: Setup Flutter
+        uses: flutter-actions/setup-flutter@v4
+        with:
+          channel: stable
+          version: latest
+          cache: true
+      - run: flutter --version
+      - run: flutter pub get
+      - name: Generate localizations
+        run: flutter gen-l10n
```
kolaente: Isn't this done automatically when running build etc?

lib/l10n/app_en.arb
```diff
@@ -0,0 +1,179 @@
+{
+  "@@locale": "en",
+  "appTitle": "Vikunja",
```
kolaente: Won't this always be the same?

lib/presentation/pages/settings_page.dart
```diff
@@ -1,270 +1,297 @@
...
+              child: Row(
+                children: [
+                  Flexible(
+                    child: TextField(
+                      inputFormatters: [FilteringTextInputFormatter.digitsOnly],
+                      keyboardType: TextInputType.number,
+                      controller: durationTextController,
+                      decoration: InputDecoration(
+                        labelText:
+                            l10n.repeatAfter, // reuse or add new key later
```
kolaente: Why the comment?

lib/presentation/widgets/task_bottom_sheet.dart
```diff
@@ -1,175 +1,202 @@
...
-        ),
-      ),
-    );
-  }
-}
+import 'package:flutter/material.dart';
+import 'package:flutter_widget_from_html/flutter_widget_from_html.dart';
+import 'package:vikunja_app/core/utils/constants.dart';
+import 'package:vikunja_app/l10n/gen/app_localizations.dart';
+// removed unused priority import after localizing strings
```
kolaente: This seems like a leftover of an AI agent :)

lib/main.dart
```diff
@@ -1,95 +1,115 @@
...
+
+  try {
+    if (!kIsWeb) {
+      final overrideCode = await settingsDatasource.getLocaleOverride();
+      final effectiveLocale = (overrideCode != null && overrideCode.isNotEmpty)
+          ? Locale(overrideCode)
+          : WidgetsBinding.instance.platformDispatcher.locale;
+      final loc = await AppLocalizations.delegate.load(effectiveLocale);
+      FileDownloader().configureNotification(
+        running: TaskNotification(loc.downloading, 'file: {filename}'),
```
kolaente: Should the `file: {filename}` part be translated too?

kolaente:
@codex review

lib/presentation/pages/settings_page.dart
RXTX4816: thanks, it was a todo note, since that key had the wrong translation
✅

.github/workflows/test.yml
RXTX4816: Apparently yes, it was a defensive approach.
I deleted those occurences, lets see how the pipeline looks after
✅

lib/l10n/app_en.arb
RXTX4816: ✅

lib/main.dart
RXTX4816: ✅

lib/presentation/widgets/task_bottom_sheet.dart
RXTX4816: ✅

lib/presentation/pages/settings_page.dart
```diff
@@ -64,8 +68,30 @@ class SettingsPageState extends ConsumerState<SettingsPage> {
                 },
               ),
             ),
+            // Language selector (override system language)
```
kolaente: this seems redundant

lib/presentation/pages/settings_page.dart
```diff
@@ -64,8 +68,29 @@ class SettingsPageState extends ConsumerState<SettingsPage> {
                 },
               ),
             ),
+            ListTile(
+              title: Text(l10n.language),
+              trailing: DropdownButton<Locale?>(
+                items: [
+                  DropdownMenuItem(
```
kolaente, line 92: I'm unsure how we could make it clear if the system language is used or not, and what is used instead.

If no language has been selected and the language is english, right now, it shows the selected language in the dialouge as "System language" but english is available as well. It's not entirely clear how the system language and English are related in this case.

kolaente: But maybe this is a non-issue, others (like Todoist) seem to have both options as well.

RXTX4816: Technically we don't need to display System Language, we can just remove it.

The benefit of having System Language is that when the user changes the system language in android settings, the app automatically adapts to it (I believe ios does it like that).

If a user selects English, his language will never change. System Language in this case is basically a way to utilize the null return value for the above mentioned convenience. 


> It's not entirely clear how the system language and English are related in this case.

I don't quite understand what you mean by that, but the only relation I see between System Language and English is that its the default locale when the Users System Language is unsupported. 

Apart from that in the best-case where the Users System Language is supported, there is no difference between selecting System Language or his language autonym directly. The former is volatile, the latter persistent. Technically no difference.

Worst case: The Users System Language is unsupported. In that case "System Language" (written in english) would be wrong, since it should say English (!= System Language). 
This case would be an issue and it would be nice for the user to have some kind of feedback/warning that the Android language is not supported, but leaving the settings dorpdown on System Language for when the language gets translated in the future.

lib/presentation/pages/settings_page.dart
RXTX4816: ✅

lib/presentation/pages/settings_page.dart
kolaente: > Technically we don't need to display System Language, we can just remove it.

Yes, but I think it would be nice if the user was able to see "we selected this language because it's your system language".

> but the only relation I see between System Language and English is that its the default locale when the Users System Language is unsupported.

That's correct, but the UI does not make that clear.

RXTX4816: Draft:
----
This message is only shown in the very specific case where the users system language is unsupported and vanishes when he selects a language.
This message would not need translations, since it would always be in english.

Screenshot From 2025-11-15 13-33-00

kolaente: I like that!

l10n.yaml
```diff
@@ -0,0 +1,8 @@
+arb-dir: lib/l10n
+template-arb-file: app_en.arb
+output-dir: lib/l10n/gen
+output-class: AppLocalizations
+output-localization-file: app_localizations.dart
+nullable-getter: false
+use-deferred-loading: false
```
leschbenedikt: Why is this flag set? It hat no effect on mobile or desktop

lib/core/utils/language_autonyms.dart
```diff
@@ -0,0 +1,22 @@
...
+String languageAutonym(Locale locale) {
+  final code = locale.languageCode.toLowerCase();
+  final country = locale.countryCode?.toUpperCase();
+
+  switch (code) {
+    case 'en':
+      return 'English';
+    case 'pl':
+      return 'Polski';
+    // On new language: Add autonyms here
```
leschbenedikt: Someone will forget to do this at some point in the future 😆 
Should we throw an error here at least in debug and add logging to force as much as possible without
interrupting production?

RXTX4816: Or we can use this: [https://pub.dev/packages/flutter_localized_locales](https://pub.dev/packages/flutter_localized_locales) and have no manual work needed when adding a new language😄

l10n.yaml
RXTX4816: Most of these flags are the default ones, I removed them

lib/core/utils/language_autonyms.dart
leschbenedikt: Saw that too. That would solve the issue.
But I'm not really a fan of "helper" dependencies. Pretty often they are not really well maintained.
The one you posted had the last commit 2 years ago so looks pretty abandoned.

Still I don't really have a better solution

RXTX4816: I was thinking the exact same as you once I saw the last commit, but when you think about it: How often do autonyms change? Once you have a set of languages and autonyms the library is done. In my opinion its a good sign and means its probably complete. It has 0 PR's and only 1 Issue, which seems good given it has 600+k downloads.

Alternative would be using another less popular package or hardcoding over 500 locale-autonym combinations.

lib/presentation/pages/settings_page.dart
RXTX4816: :heavy_check_mark:

RXTX4816:
Ready to review.
I added tests that check the .arb files against the generated locale files and prints a warning message if and how many untranslated strings there are for each locale. Its just a warning, no error.
This might come in handy in the future.

The log output (from this pipeline) looks like this:
```
/home/runner/work/app/app/test/localization_test.dart: Localization coverage Count untranslated strings per supported locale (vs English baseline)
  Locale es: 158 untranslated strings
  Locale pl: 3 untranslated strings
```

lib/core/utils/language_autonyms.dart
leschbenedikt: What I've learned so far is that you should have as little dependencies as possible.

It's not about if it's working or not.
There could be other more severe issues.
Like what happens if there is a security issue. Or when it's deleted.
What will we do if it's not compatible anymore. 

There are a lot of things that could happen.
In such case we could be block for several days.
And for a production env with several user this is always bad.

It's always a trade of to have dependencies.
But in this case I don't think the benefit is worth the cost adding a dependencies on another project.

In fact the lib is just a list of locale-autonym combinations.
Why shouldn't we do the same. It's not that much work and we are safe.

RXTX4816: Good point you make there. 

I removed the 3rd party library.

lib/core/utils/language_autonyms.dart
```diff
@@ -0,0 +1,83 @@
+import 'package:flutter/widgets.dart';
+import 'package:intl/intl.dart';
+
+const Map<String, String> kLocaleAutonyms = <String, String>{
```
kolaente, line 4: Can we add this only when adding new languages? That would match the process in the frontend.

RXTX4816: Sure, I removed all but english and polish :+1:

Issue #164, opened 2025-11-04
Exception: type 'String' is not a subtype of type 'Map<String, dynamic>'


csd4ni3l:
I get this exception in the Vikunja app on Android 13 in the Projects and Settings tab. I saw similar issues, but not this exactly. 

Vikunja server version is v0.24.6

![Image](https://github.com/user-attachments/assets/9903b600-05dd-49d2-8428-089ad611f033)

leschbenedikt:
The issue is that you are on an older server version.
The app is currently only optimized for the newest api of the newest server version.

If you update your server the issue should be gone.

In the future there maybe could be a way to support multiple server versions

csd4ni3l:
Good to know, do you mean i should use the latest commit? I am gonna try a bit later and will report if it works

leschbenedikt:
Yes, the newest version should be:

https://github.com/go-vikunja/vikunja/releases/tag/v1.0.0-rc2

csd4ni3l:
Yes, it works now. Thanks.

PR #179, opened 2025-11-13
fix: ignore filter if someone with outdated server is using the app

leschbenedikt:


RXTX4816 (approved):
Looks good. 
Original issue #164 can be reproduced with old server version.

Your proposed change solves the issue and all Projects are shown in the Projects view, even if they were created on an older server version.

PR #180, opened 2025-11-13
fix: wrong cast

leschbenedikt:


RXTX4816 (approved):
Looks good and is in-line with [flutter's approach](https://api.flutter.dev/flutter/dart-async/AsyncError/defaultStackTrace.html)

kolaente (approved):

PR #182, opened 2025-11-14
feat: login URL without protocol treated as https

### LINKED ISSUE #99: If the Vikunja api protocol is not present, try different protocols
When the user enters a url like `example.com` instead of `https://example.com`, the app will try to make requests regardless and fail.

Relevant issue in Sentry: https://vikunja.sentry.io/issues/5938276808/ (not public because the Vikunja url is in there).

### LINKED ISSUE #108: ArgumentError: Invalid argument(s): Missing scheme in uri: vikunja.xxxx.ts.net
Sentry Issue: [APP-D](https://vikunja.sentry.io/issues/5938276808/?referrer=github_integration)

```
ArgumentError: Invalid argument(s): Missing scheme in uri: vikunja.xxxx.ts.net
  File "webview_controller.dart", line 173, in WebViewController.loadRequest
  File "login_webview.dart", line 32, in LoginWithWebViewState.initState
...
(138 additional frame(s) were not displayed)
```

It looks like there should be a validation when entering the url - or even better, add `http(s)://` automagically when entering the url.

RXTX4816:
resolves #99 
resolves #108

### DIFF
--- a/lib/core/utils/validator.dart
+++ b/lib/core/utils/validator.dart
@@ -16,9 +16,9 @@ bool isUrl(String? url) {
   return _url.hasMatch(url);
 }
 
-String? isURLValid(String? url) {
-  if (url == null || url.isEmpty) return null;
+bool isURLValid(String? url) {
+  if (url == null || url.isEmpty) return true;
   final trimmed = url.trim();
-  if (isUrl(trimmed) || isUrl('https://$trimmed')) return null;
-  return 'Invalid URL';
+  if (isUrl(trimmed) || isUrl('https://$trimmed')) return true;
+  return false;
 }
--- a/lib/presentation/pages/login/login_page.dart
+++ b/lib/presentation/pages/login/login_page.dart
@@ -212,7 +212,7 @@ class LoginPageState extends ConsumerState<LoginPage> {
             focusNode: focusnode,
             enabled: !_loading,
             validator: (address) {
-              return isURLValid(address);
+              return isURLValid(address) ? null : 'Invalid URL';
             },
             decoration: InputDecoration(
               border: OutlineInputBorder(),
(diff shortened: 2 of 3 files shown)


kolaente:
There are many whitespace changes on this, are you on Windows?

lib/presentation/pages/login/register_page.dart
```diff
@@ -1,168 +1,185 @@
...
+      );
+    } finally {
+      setState(() {
+        _loading = false;
+      });
+    }
+  }
+}
+
+String _normalizeServer(String input) {
```
kolaente: Why does this function exist twice?

RXTX4816: Good catch 😉
Fixed ✅

RXTX4816:
> There are many whitespace changes on this, are you on Windows?
Update: VSCode on Linux did me dirty with its default "EndOfLine=auto" feature.. fixed.

lib/presentation/pages/login/login_page.dart
```diff
@@ -1,443 +1,453 @@
...
+    } catch (ex) {
+      _showGenericError(context);
+    } finally {
+      setState(() {
+        _loading = false;
+      });
+    }
+  }
+
+  String _normalizeServer(String input) {
```
kolaente: This should use the helper function you just extracted

lib/presentation/pages/login/register_page.dart
kolaente: and the function still exists here?

lib/presentation/pages/login/login_page.dart
RXTX4816: My apologies, there were technical difficulties with git on my end. Fixed now!

lib/core/utils/validator.dart
```diff
@@ -15,3 +15,10 @@ bool isUrl(String? url) {
   if (url == null) return false;
   return _url.hasMatch(url);
 }
+
+String? isURLValid(String? url) {
+  if (url == null || url.isEmpty) return null;
+  final trimmed = url.trim();
+  if (isUrl(trimmed) || isUrl('https://$trimmed')) return null;
+  return 'Invalid URL';
```
kolaente: Since this is a utility function now, please return a bool whether the URL is valid or empty instead of returning an error message.

RXTX4816: ✅

kolaente (approved):

Issue #173, opened 2025-11-10
Task sorting in task overview and project detail


leschbenedikt:
As of now the sorting of task in the task overview and project detail doesn't seem to be consistent, especially after adding a new task.

It should match the sorting of the we application

PR #183, opened 2025-11-15
fix: task sorting in home and listview

RXTX4816:
#173

fix: task order is according to web app
fix: kanban view show no 'Null' type error
fix: "show done tasks" in project list view
feat: tasks are sorted on new task
feat: tasks in project list view are draggable

### DIFF
--- a/lib/presentation/manager/project_controller.dart
+++ b/lib/presentation/manager/project_controller.dart
@@ -25,7 +25,7 @@ class ProjectController extends _$ProjectController {
       case SuccessResponse<List<Task>>():
         return ProjectPageModel(
           project,
-          -1,
+          0,
           tasksResponse.body,
           [],
           displayDoneTask,
--- a/lib/presentation/widgets/project/project_task_list.dart
+++ b/lib/presentation/widgets/project/project_task_list.dart
@@ -163,8 +163,8 @@ class ProjectTaskList extends ConsumerWidget {
     return ProjectTaskListItem(
       key: Key(task.id.toString()),
       task: task,
-      onTap: () => _showTaskBottomSheet(ref.context, task),
-      onEdit: () => _onEdit(ref.context, task),
+      onTap: () => _showTaskBottomSheet(ref, task),
+      onEdit: () => _onEdit(ref, task),
       onCheckedChanged: (value) async {
         var success = await ref
             .read(projectControllerProvider(project).notifier)
@@ -180,26 +180,27 @@ class ProjectTaskList extends ConsumerWidget {
     );
   }
 
-  void _showTaskBottomSheet(BuildContext context, Task task) {
+  void _showTaskBottomSheet(WidgetRef ref, Task task) {
     showModalBottomSheet<void>(
-      context: context,
+      context: ref.context,
       shape: RoundedRectangleBorder(
         borderRadius: BorderRadius.vertical(top: Radius.circular(10.0)),
       ),
       builder: (BuildContext context) {
-        return TaskBottomSheet(
-          task: task,
-          onEdit: () => _onEdit(context, task),
-        );
+        return TaskBottomSheet(task: task, onEdit: () => _onEdit(ref, task));
       },
     );
   }
 
-  void _onEdit(BuildContext context, Task task) {
-    Navigator.push<Task>(
-      context,
+  void _onEdit(WidgetRef ref, Task task) async {
+    var editedTask = await Navigator.push<Task?>(
+      ref.context,
       MaterialPageRoute(builder: (buildContext) => TaskEditPage(task: task)),
     );
+
+    if (editedTask != null) {
+      ref.read(projectControllerProvider(project).notifier).reload();
+    }
   }
 
   void _navigateToDetail(BuildContext context, Project project) {
--- a/lib/core/di/network_provider.dart
+++ b/lib/core/di/network_provider.dart
@@ -11,7 +11,10 @@ class AuthData extends _$AuthData {
   @override
   AuthModel? build() => null;
 
-  void set(AuthModel token) => state = token;
+  void set(AuthModel token) {
+    state = token;
+    ref.invalidate(clientProviderProvider);
+  }
 }
 
 @Riverpod(keepAlive: true)
@@ -22,9 +25,12 @@ class CurrentUser extends _$CurrentUser {
   void set(User user) => state = user;
 }
 
-@riverpod
-Client clientProvider(Ref ref) {
-  final authData = ref.watch(authDataProvider);
+@Riverpod(keepAlive: true)
+class ClientProvider extends _$ClientProvider {
+  @override
+  Client build() {
+    final authData = ref.read(authDataProvider);
 
-  return Client(base: authData?.address ?? '', token: authData?.token);
+    return Client(base: authData?.address ?? '', token: authData?.token);
+  }
 }
--- a/lib/core/di/network_provider.g.dart
+++ b/lib/core/di/network_provider.g.dart
@@ -6,24 +6,7 @@ part of 'network_provider.dart';
 // RiverpodGenerator
 // **************************************************************************
 
-String _$clientProviderHash() => r'5b68f6922e1223e1b0a4d106a013c69df0c052c6';
-
-/// See also [clientProvider].
-@ProviderFor(clientProvider)
-final clientProviderProvider = AutoDisposeProvider<Client>.internal(
-  clientProvider,
-  name: r'clientProviderProvider',
-  debugGetCreateSourceHash: const bool.fromEnvironment('dart.vm.product')
-      ? null
-      : _$clientProviderHash,
-  dependencies: null,
-  allTransitiveDependencies: null,
-);
-
-@Deprecated('Will be removed in 3.0. Use Ref instead')
-// ignore: unused_element
-typedef ClientProviderRef = AutoDisposeProviderRef<Client>;
-String _$authDataHash() => r'37d6f0a9be23c0f1bbe648fe223fbedcd22002e5';
+String _$authDataHash() => r'321e34c6d952287a0371852203f99572aa89a741';
 
 /// See also [AuthData].
 @ProviderFor(AuthData)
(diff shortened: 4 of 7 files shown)


lib/presentation/manager/project_controller.dart
```diff
@@ -77,22 +88,82 @@ class ProjectController extends _$ProjectController {
 
   Future<Response<List<Task>>> _loadTasks(
     int projectId,
-    bool displayDoneTasks,
-  ) async {
+    bool displayDoneTasks, {
+    int? viewId,
```
leschbenedikt: What the use case of having a nullable viewId?
Why do we need to load something without view?

lib/presentation/manager/project_controller.dart
```diff
@@ -77,22 +88,82 @@ class ProjectController extends _$ProjectController {
...
+    };
+
+    if (viewId != null) {
+      if (!displayDoneTasks) {
+        final params = {
+          ...baseParams,
+          "filter": ["done=false"],
+        };
+        return repo.getAllByProjectView(projectId, viewId, params);
+      } else {
```
leschbenedikt: This else case seems to be overcomplicated.
Is this not just the above without filter?

lib/presentation/manager/project_controller.dart
```diff
@@ -77,22 +88,82 @@ class ProjectController extends _$ProjectController {
...
+          viewId,
+          paramsDoneView,
+        );
+        List<Task> doneTasks;
+        if (doneRespView.isSuccessful &&
+            doneRespView.toSuccess().body.isNotEmpty) {
+          doneTasks = sortTasksByPosition(doneRespView.toSuccess().body);
+        } else {
+          // Fallback: project-level done tasks
+          final paramsDoneProject = {
```
leschbenedikt: I don't think a fallback makes sense.
If I can't load what I want, I want it to fail not to show different data.
And if the data is the same this is just a retry which should  be handled generally

lib/presentation/manager/project_controller.dart
```diff
@@ -42,11 +49,15 @@ class ProjectController extends _$ProjectController {
         .getDisplayDoneTasks(project.id);
 
     var tasks = <Task>[];
-    var tasksResponse = await _loadTasks(project.id, displayDoneTask);
+    var tasksResponse = await _loadTasks(
```
leschbenedikt: This will have wrong task data in kanban view.
The kanban view response looks different that the others.
That why it was done like this.

I'm not fully convinced by the old version either.

lib/presentation/widgets/project/project_task_list.dart
```diff
@@ -22,6 +22,86 @@ class ProjectTaskList extends ConsumerWidget {
 
     return projectController.when(
       data: (pageModel) {
+        if (project.subprojects.isEmpty && pageModel.tasks.isNotEmpty) {
+          final openTasks = pageModel.tasks.where((t) => !t.done).toList();
+          final doneTasks = pageModel.tasks.where((t) => t.done).toList();
+          return CustomScrollView(
+            slivers: [
+              SliverReorderableList(
```
leschbenedikt: Why am I only allowed to sort open task? The web uI allows both.
Also open and done task are not separated from each other.

When we combine the and allow sorting of all we can also reduce the complexity

lib/presentation/manager/project_controller.dart
```diff
@@ -277,9 +416,19 @@ class ProjectController extends _$ProjectController {
 
     var value = state.value;
     if (value != null) {
-      var tasksResponse = await _loadTasks(project.id, displayDoneTasks);
+      final currentViewId =
+          value.viewIndex >= 0 && value.viewIndex < value.project.views.length
```
leschbenedikt: Can this ever happen?

lib/core/utils/sort_tasks.dart
```diff
@@ -0,0 +1,32 @@
+import 'package:vikunja_app/domain/entities/task.dart';
+
+List<Task> sortTasksByPosition(List<Task> tasks) {
```
leschbenedikt: Can we do this by passing a sorting params to the backend?
This is how it was handled in other cases

lib/core/utils/sort_tasks.dart
```diff
@@ -0,0 +1,32 @@
...
+    if (aHas != bHas) {
+      return aHas ? -1 : 1;
+    }
+
+    return b.id.compareTo(a.id);
+  });
+  return sorted;
+}
+
+List<Task> sortTasksByDoneThenPosition(List<Task> tasks) {
```
leschbenedikt: You don't need to separate them.  It easier to make it a sort criteria

lib/data/models/task_dto.dart
```diff
@@ -55,21 +55,29 @@ class TaskDto extends Dto<Task> {
   TaskDto.fromJson(Map<String, dynamic> json)
     : id = json['id'],
       title = json['title'],
-      description = json['description'],
-      identifier = json['identifier'],
-      done = json['done'],
+      description = json['description'] ?? '',
```
leschbenedikt: You don't need all of this checks. That's done by the backend for us

lib/presentation/manager/project_controller.dart
```diff
@@ -201,6 +272,74 @@ class ProjectController extends _$ProjectController {
...
+
+    final currentViewIndex = value.viewIndex >= 0 ? value.viewIndex : 0;
+    if (project.views.isEmpty ||
+        project.views[currentViewIndex].viewKind != ViewKind.list) {
+      return false;
+    }
+
+    if (oldIndex == newIndex) return true;
+
+    final tasks = [...value.tasks];
```
leschbenedikt: Aren't most of the check already done in the onReorder before?
Also I think the logic should be all in one place. At best in the list.

Additionally I think we can reduce the complexity of this a lot if we just allow sorting of all items like in the web ui.

To be honest I don't want to be the one debugging and maintaining this 😄

lib/presentation/manager/project_controller.dart
```diff
@@ -201,6 +272,74 @@ class ProjectController extends _$ProjectController {
...
+    final nextIndex = targetIndex + 1;
+    double? after;
+    if (moved.done) {
+      after = (nextIndex < tasks.length) ? tasks[nextIndex].position : null;
+    } else {
+      after = (nextIndex < openBoundary) ? tasks[nextIndex].position : null;
+    }
+
+    double newPos;
+    if (before != null && after != null) {
```
leschbenedikt: There is a special logic for calculating the new position. Look at `calculateItemPosition`

lib/presentation/manager/project_controller.dart
RXTX4816: Hello Benedikt, thanks for your review :) Yeah makes sense to have them all inter-sortable like in the web. I will try to implement these changes in the following days, currently time is short.

lib/core/utils/sort_tasks.dart
RXTX4816: Currently not, position as sort_by needs a viewid, but the API Url we currently use for this is undocumented, and definitly does not take viewid. 
There are a view different API Endpoints in the docu that specifically allow "position" as sort_by, although they dont implement "viewid" passing. This will error on the Backend. 

Concluding: Before anything is done here, first we need to be sure that the API is consistent across the APP and the BACKEND and that the documentation is correct. Otherwise it makes no sense. 

In my opinion the API client of the app must first be implemented to respect every endpoint stated here [https://try.vikunja.io/api/v1/docs#tag/task](https://try.vikunja.io/api/v1/docs#tag/task)

I will try to implement a new API call for the ListView

lib/presentation/manager/project_controller.dart
RXTX4816: If you call _loadTasks with viewId null, it uses the previous implementation. 
Currently this is necessary when calling from kanban.

In the future it should be refactored.

lib/presentation/manager/project_controller.dart
RXTX4816: Yes, refactored.
Now: Uses either the old filters or a new API call with the position sort_by filter.

lib/presentation/manager/project_controller.dart
RXTX4816: Good catch, this is fatally flawed.
Its deleted now with the refactor

lib/presentation/manager/project_controller.dart
RXTX4816: Yeah its really something, the implementation there.. 

Anyway, with the viewId null it just uses the old way, so nothing will break for kanban.

I tested this with a project that has only kanban view, and the current solution works, since I didn't change legacy code.

One can now still just call _loadTasks(id, done) without giving viewid.

lib/presentation/widgets/project/project_task_list.dart
RXTX4816: Yes good idea, I implemented it :+1:

lib/presentation/manager/project_controller.dart
RXTX4816: It was a bug, I thought viewIndex was viewId. 
Fixed now.

lib/core/utils/sort_tasks.dart
RXTX4816: /projects/{id}/views/{view}/tasks
	
Now uses the correct API call for getting the tasks of a project.

But only if the List view is opened from another view, or the first view of a project is the List view and that project is opened, or a reload occurs from within the list view.

For the other views, the old (undocumented) api call is used.

lib/core/utils/sort_tasks.dart
RXTX4816: It was not easier, that I can say, but its a more elegant solution with sort_by(position) :)

lib/data/models/task_dto.dart
RXTX4816: Yeah I removed them, that was not to touch anyway for this MR

lib/presentation/manager/project_controller.dart
RXTX4816: You are right, I looked at the moveTask structure, but they use (maybe more updated?) widgets and an own directory for the ui stuff.
But I moved the compute logic to the same file where the list is.

lib/presentation/manager/project_controller.dart
RXTX4816: Yes yes I did not notice :laughing: 
The solution uses it now.

lib/presentation/widgets/project/project_task_list.dart
```diff
@@ -22,6 +23,66 @@ class ProjectTaskList extends ConsumerWidget {
 
     return projectController.when(
       data: (pageModel) {
+        if (project.subprojects.isEmpty && pageModel.tasks.isNotEmpty) {
```
leschbenedikt: Why is the feature only available if there is no subproject

lib/presentation/widgets/project/project_task_list.dart
```diff
@@ -22,6 +23,66 @@ class ProjectTaskList extends ConsumerWidget {
...
+                    if (context.mounted && !success) {
+                      ScaffoldMessenger.of(context).showSnackBar(
+                        const SnackBar(content: Text('Failed to reorder task')),
+                      );
+                    }
+                  });
+            },
+            itemBuilder: (context, index) {
+              final task = pageModel.tasks[index];
+              return ReorderableDelayedDragStartListener(
```
leschbenedikt: The dividers are missing

lib/presentation/widgets/project/project_task_list.dart
RXTX4816: I integrated the new Sliver stuff into the old , fixed

lib/presentation/widgets/project/project_task_list.dart
RXTX4816: Fixed

PR #186, opened 2025-11-16
feature: more intuitive task description editing

### LINKED ISSUE #175: More intuitive task edition
At the moment it is not very intuitive how to edit the description of a task. 
You need to click on the text or icon but not in-between or anywhere else in the row.
Even empty line between text does not work for multi line text.

It should be more intuitive for the user to get to the description editor

leschbenedikt:
Resolves #175 

Added a label to make it easier to find where the description actually is and made the hole row clickable.

image

Also adjusted the paddings slightly to make it more consistant


kolaente (approved):
Thanks!

kolaente:
Not sure why the build fails, this seems unrelated.

PR #190, opened 2025-11-25
feat: add task comments support with dedicated edit page

### LINKED ISSUE #115: "Add comment" for tasks
Commenting on tasks is by far one of my most frequent actions on the desktop client. 
Would love for this feature to be one the app, since I often think of things to add while on the go.

13rac1:
Add ability to view, create, edit, and delete comments on tasks. My intention is to create a minimum implementation of the feature to reduce code review time. I'd like to see comments on the task preview too, but that can happen later.

Changes:
- New TaskComment entity and DTO with JSON serialization
- TaskCommentDataSource for API communication (GET/PUT/POST/DELETE)
- TaskCommentRepository interface and implementation
- TaskCommentsController for state management (Riverpod with code generation)
- NEW: Dedicated CommentEditPage with HtmlEditor for rich text editing
- TaskComments widget displaying comments with HTML rendering
- Comments section added to task edit page
- NEW: Delete confirmation dialog for comments
- ~~Move save button from FAB to app bar to avoid overlap with comment input - This is a UX change. I tried to avoid it to reduce the changeset and avoid the distraction, but nothing else makes sense. I'm open to other options.~~
- Add TaskCommentDto tests for JSON parsing

Notice: This code was primarily generated by Claude Code, but I have reviewed and tested it; it's mostly simple CRUD. There's multiple "nice to haves" which could be implemented, but nothing seems like a blocker to me. Please feel free to rip it apart! 

Fixes #115

No Comments
Comment Added

## Release Notes

* **New Features**
  * Added task comments functionality—users can now view, add, edit, and delete comments on task detail pages with rich text editing support and comment history including author and timestamp information.

* **Tests**
  * Added comprehensive unit test coverage for task comments data handling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


RXTX4816:
Hi Cool Feature! I tested it and it works very good. 

I did not review the code, but have some notes regarding the UI/UX design:

With your proposed design, from UX perspective, the User could mistakenly save his input instead of deleting it, when the buttons are that close next to each other and saving has no confirmation.

Previously, with the Save Button in the bottom right, and e.g. having Attachements in a Task and Editing it, the Download Button also overlaps with the Save Button. But since "Scrolling Down" shows like half the screen of empty space, this is a good workaround. In your solution, when scrolling to the bottom, one cannot scroll into empty space, so it definetly makes sense to have the Save Button moved. It should be discussed how such a UX decision is handled. 
<details>
  <summary>Empty Space Example Screenshot</summary>
  

IMG_20251205_233142
  
</details>


I think the "scrolling into empty space" solution is better from a UX perspective and for less thumb-stretching, but lets see what the others have to say :smiley:

13rac1:
@RXTX4816 Thanks for the review!

Here's the problem with the FAB in that location: 
image
The text field is supposed to be directly above the keyboard during text entry, but now the FAB is on top of the submit button, which breaks the regular Android text entry UX. IMO The FAB isn't necessary on the Task Edit page at all, the Android app _should_ auto-save like the web app does. That being said, I'm open to implementing whatever UX just to get the comment CRUD on this app so I can stop using the web app on my phone. 

Some UX options:
1. Remove the FAB and put the Save button in the App Bar menu (current PR)
2. Hide the FAB when the comment entry text field is focused.
3. Move the comment create/edit to an entirely separate full screen view (this is how Github app comments work)
4. Move the FAB upwards during comment entry. **Warning**: There will be a Comment submit arrow button and a FAB on top of each other.

I'm open to 1, 2, and 3. I don't like the UX of 2, but even it is acceptable for now. 

@kolaente I'm open to whatever UX you prefer. Or, if you don't like that a LLM wrote the CRUD, I'll close the PR. LMK thanks!

RXTX4816:
@13rac1 Thanks for the screenshot, I now see what you mean and its indeed tricky. 

Regarding the auto-saving you mentioned I tested it again and you should not (IMO) be able to Add, Edit or Delete comments immediatly, at least not from the task edit page, which should only apply changes if the user clicks on "Save". Personally I think it is great to add the comments fast like that, but it also makes the Task Edit Page incosistent.
On Android handhelds I believe its more common to not have auto-save to backend like that, especially in production ready environments with User Management and Rights Access. 
One idea that came in mind is using the "Task Overview"(=The one that shows when left clicking on any Task) Screen for quick comment adding + viewing and leaving the editing + deleting in the Task Edit Page, or something like that. 
Or maybe having a seperate Comments modal dialog would also solve these issues(Also the UIUX one) while not slowing down the users workflow unnecessarly. The modal dialog would allow immediate (=auto-save) comment operations without the need for a save button and could be called from e.g. the Task Edit Page or the Tasks Overview Page. 
This would be your UX Point 3. 

Regarding the save button:
Another option would be to shrink the size of the comments section on the left and right side, like the Attachements are displayed in the Task Edit Page. This would ensure both Submit and Save buttons are always visible (With slight overlap like the Download button). But yeah I think you listed all the available UIUX options.

But again, lets see what the product owner has to say :laughing:

kolaente:
Hey thanks for your thoughts on this!

I think having a separate screen just to comment would be the best option here, the third option proposed by @13rac1
This would also allow using the full power of the editor to format the comment.

13rac1:
@kolaente @RXTX4816 I've refactored to use a separate Create/Edit comment view. Again, this is the absolute minimum implementation. Confirmation dialogs, different formatting, and a fancy HTML/markdown text editor can be added later, but will bloat the PR now.

image
image

RXTX4816:
Thanks for the changes, maybe there was a missunderstanding from my side, I thought UX Point 3 was to create a completely seperate screen (not in task edit screen), to directly manipulate comments, but this is fine also but I have some notes after I tested the recent changes:

- The Add and Edit comment uses a TextBox that returns a String. This is in contrast to how the Webapp creates comments, where every comment is by default HTML wrapped in a `<p>` html tag. Editing a comment created through the Webapp currently shows the `<p>` tag visibly, while in the webapp its not shown visibly. Similar implementation for HTML Textbox can be found in _buildDescription. :warning: 
- Adding and Editing a comment is immediate, although the Task Edit page should hold all changes in memory until the User presses the "Save" Button on the bottom right.  
- Deleting is also immediate, the "Confirm" dialog should not be necessary in the Task Edit Screen since the changes are in memory until user exits the screen or saves = async http call)

Why is this a problem from UX point of view? Lets say I edit multiple fields of a task, the changes are in memory. Now I add a comment and maybe delete another comment. I now press "Return" (I want to omit the current changes), I'm asked "Would you like to dismiss those changes?" I click "Dismiss". But the comment I added already went through, so it is not possible to dismiss that action, which would be a contradiction.

Thats why I thought there could be a totally seperate screen opposed to Task Edit like Comment Edit that applies changes instantly. But since you want to keep it minimal, maybe it should be left in the Task Edit page and have it not auto-apply changes directly. 
The HTML editor would be necessary to ensure compatible behaviour with Webapp.

Update: UIUX will be in follow up PR, for this PR just the Textbox and its fine from my side 👍 Thanks for the work

lib/presentation/pages/task/comment_edit_page.dart
```diff
@@ -0,0 +1,108 @@
...
+                    child: CircularProgressIndicator(strokeWidth: 2),
+                  )
+                : const Icon(Icons.save),
+            onPressed: _isValid && !_isSaving ? _save : null,
+          ),
+        ],
+      ),
+      body: Padding(
+        padding: const EdgeInsets.all(16.0),
+        child: TextField(
```
kolaente: Please replace this `TextField` with a `HtmlEditor` as is the case for the description edit screen. This will give us HTML editing of comments without much effort.

kolaente:
I think having a separate edit screen for comments is fine for now, we might move the whole thing (view and edit) over to a completely separate screen in a follow-up PR when we notice that's what's needed.

lib/presentation/pages/task/comment_edit_page.dart
13rac1: Done!

13rac1:
I'm looking forward to comments in any form 😂 

@RXTX4816 I continue to prefer auto-save of everything like the webapp. In this case though, creating/updating comments at the same time is far less work. Storing comment changes in memory will require keeping changeset history with transaction log, making multiple API calls, and probably _retrying_ HTTP requests as needed to ensure consistency. It's far more complex, but could be done!

RXTX4816:
@13rac1 I did not think about that, appreciate the notes. 
No comments from my side, code looks good and the feature works correctly. 
Once its merged I will open an Issue about the UI/UX and summarize the notes from here :+1:

kolaente:
@coderabbitai full review

lib/data/models/task_comment_dto.dart
```diff
@@ -0,0 +1,52 @@
...
+    DateTime? updated,
+  }) : created = created ?? DateTime.now(),
+       updated = updated ?? DateTime.now();
+
+  TaskCommentDto.fromJson(Map<String, dynamic> json)
+    : id = json['id'] ?? 0,
+      comment = json['comment'] ?? '',
+      author = UserDto.fromJson(json['author']),
+      created = DateTime.parse(json['created']),
+      updated = DateTime.parse(json['updated']);
```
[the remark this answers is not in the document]
kolaente: This is not relevant.

lib/domain/entities/task_comment.dart
```diff
@@ -0,0 +1,18 @@
...
+  final DateTime updated;
+
+  TaskComment({
+    this.id = 0,
+    required this.comment,
+    required this.author,
+    DateTime? created,
+    DateTime? updated,
+  }) : created = created ?? DateTime.now(),
+       updated = updated ?? DateTime.now();
```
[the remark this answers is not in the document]
kolaente: This is not relevant.

lib/presentation/pages/task/comment_edit_page.dart
```diff
@@ -0,0 +1,93 @@
...
+
+    final text = await _controller.getText();
+
+    // Check if text is empty after getting it
+    if (text.trim().isEmpty) {
+      setState(() => _isSaving = false);
+      ScaffoldMessenger.of(
+        context,
+      ).showSnackBar(const SnackBar(content: Text('Comment cannot be empty')));
+      return;
```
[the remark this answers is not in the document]
kolaente: @13rac1 please implement this (and similar issues)

lib/presentation/widgets/task/task_comments.dart
```diff
@@ -0,0 +1,186 @@
...
+      );
+    }
+
+    return ListView.builder(
+      shrinkWrap: true,
+      physics: const NeverScrollableScrollPhysics(),
+      itemCount: comments.length,
+      itemBuilder: (context, index) =>
+          _buildCommentItem(context, ref, comments[index]),
+    );
```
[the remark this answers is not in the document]
kolaente: This is fine.

kolaente:
@13rac1 please look at the unresolved comments from coderabbit, from my understanding, the points it raised are valid.

If you don't think using AI to review PRs is a good idea or devalues your work, please tell me! I've just seen Claude in your commit messages so I assume you're using coding agents as well and this is fine.

13rac1:
@kolaente No problem at all! My priority is :shipit:

13rac1:
@kolaente I've updated code. I confirmed each change and 👍🏼 to what the 🤖 🐰 asked.

test/task_comment_dto_test.dart
```diff
@@ -0,0 +1,163 @@
...
+
+      expect(comment.id, 42);
+      expect(comment.comment, 'This is a test comment');
+      expect(comment.author.id, 1);
+      expect(comment.author.username, 'testuser');
+      expect(comment.created, DateTime.utc(2024, 2, 1, 9, 0, 0));
+      expect(comment.updated, DateTime.utc(2024, 2, 1, 9, 5, 0));
+    });
+
+    test('parses comment with missing optional id', () {
```
kolaente: Tasks coming from the api will always have an ID, so this seems a little redundant (the test and allowing to parse tasks that don't have an ID)

lib/presentation/widgets/task/task_comments.dart
```diff
@@ -0,0 +1,191 @@
...
+    final commentsAsync = ref.watch(taskCommentsControllerProvider(taskId));
+
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: [
+        Padding(
+          padding: const EdgeInsets.symmetric(vertical: 8.0),
+          child: Row(
+            children: [
+              Text('Comments', style: Theme.of(context).textTheme.titleMedium),
```
kolaente: This needs a translation

lib/presentation/widgets/task/task_comments.dart
```diff
@@ -0,0 +1,191 @@
...
+                tooltip: AppLocalizations.of(context).addCommentTooltip,
+              ),
+            ],
+          ),
+        ),
+        commentsAsync.when(
+          data: (comments) => _buildCommentsList(context, ref, comments),
+          loading: () => const Center(child: CircularProgressIndicator()),
+          error: (error, _) => const Text(
+            'Failed to load comments',
```
kolaente: This needs a translation as well

lib/presentation/widgets/task/task_comments.dart
```diff
@@ -0,0 +1,191 @@
...
+
+  Widget _buildCommentsList(
+    BuildContext context,
+    WidgetRef ref,
+    List<TaskComment> comments,
+  ) {
+    if (comments.isEmpty) {
+      return const Padding(
+        padding: EdgeInsets.symmetric(vertical: 8.0),
+        child: Text('No comments yet', style: TextStyle(color: Colors.grey)),
```
kolaente: and here as well

lib/presentation/widgets/task/task_comments.dart
```diff
@@ -0,0 +1,191 @@
...
+                        onSelected: (value) {
+                          if (value == 'edit') {
+                            _navigateToEditPage(context, taskId, comment);
+                          } else if (value == 'delete') {
+                            _deleteComment(context, ref, taskId, comment);
+                          }
+                        },
+                        itemBuilder: (context) => const [
+                          PopupMenuItem(value: 'edit', child: Text('Edit')),
+                          PopupMenuItem(value: 'delete', child: Text('Delete')),
```
kolaente: same here

lib/presentation/widgets/task/task_comments.dart
```diff
@@ -0,0 +1,191 @@
...
+      child: Padding(
+        padding: const EdgeInsets.all(12),
+        child: Column(
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            Row(
+              mainAxisAlignment: MainAxisAlignment.spaceBetween,
+              children: [
+                Text(
+                  comment.author.name.isNotEmpty
```
kolaente: This should be refactored in a central method, we'll need it in other places as well.

lib/presentation/widgets/task/task_comments.dart
```diff
@@ -0,0 +1,191 @@
...
+                      ? comment.author.name
+                      : comment.author.username,
+                  style: const TextStyle(fontWeight: FontWeight.bold),
+                ),
+                Row(
+                  children: [
+                    Text(
+                      dateFormat.format(comment.created.toLocal()),
```
kolaente: The web ui shows the edited date as well, if the comment was edited. Please do that here as well.

test/task_comment_dto_test.dart
13rac1: Removed

lib/presentation/widgets/task/task_comments.dart
13rac1: Moved to an extension file

lib/presentation/widgets/task/task_comments.dart
13rac1: Both dates now show

13rac1:
Made the requested changes in a separate commit for clarity

PR #197, opened 2025-12-09
feat(i18n): add Italian localization file

catoblepa:


RXTX4816:
Could you also add the autonym to language_autonyms.dart? 

We should probably mention that step in the contributions section of the readme.

kolaente:
Translations are now synced automatically from CrowdIn. 

I've uploaded the changes from this PR here to Crowdin so that the Italian translation is already present there, please head over to Crowdin to complete the translation. Also check out the translation docs: https://vikunja.io/docs/translations/

Please modify this PR so that it only adds Italian for selection in the settings.

PR #200, opened 2025-12-10
fix: Sharing multiple lines of text only shows the first but adds all…

### LINKED ISSUE #178: Sharing multiple lines of text only shows the first but adds all lines
Follow-up from https://github.com/go-vikunja/app/pull/176:

> I noticed that when sharing multiple lines of text, only the first line would be shown in the add dialogue but all lines will be added as the task title.

leschbenedikt:
Resolves #178


kolaente (approved):
LGTM, still draft?

PR #201, opened 2025-12-10
fix: ignore bucket filter for outdated server versions

leschbenedikt:


kolaente:
The changes look good, but I think maybe we should just not allow using the app with an outdated server version at all? Because this will also affect other things.

leschbenedikt:
That depends on how api changes are handled in the future.

If there will be breaking changes in the existing apis we should block other server versions.

if changes are versioned we can allow older versions until they are no longer supported.

For now I just wanted to make the app more stable

kolaente:
There might be breaking changes as covered by Semver - we will make a new major release in that case.

To make this more maintainable overall, I would only ever support the latest major release.

leschbenedikt:
There are basically 3 possibilities:

- Server version is supported by client 
  -> Here we do not have a problem
- Server version is older than the supported version 
  -> We can block here. It's not optimal from a user perspective but it is easy to solve by updating the server.
- Server version is newer than the supported version 
  -> Here we could have a problem. It could probably take some time until the client is adapted. Especially if you wait for a     stable release. Until this the client could be broken. How should we handle this?

kolaente:
Maybe for each major version we can support the clients without code changes? For breaking changes we'll have a new Major release and that would also (ideally) be coordinated it with the clients.

The current situation is a bit special because we didn't have major releases yet, but v1.0.0 s right around the corner.

leschbenedikt:
We also could have a problem with the first case when there is no server release as in #204 

Normally I would solve that topic by versioning the API. That would mean that the server and the client need to support multiple version of a endpoint for a short time. But this would solve it entirely.

If we don't want to support multiple versions on the server we could also check for the server version and decide what to do based on that. This could solve some issues similar to this PR or at least wouldn't crash the app.

kolaente:
I'm just really hesitant to support multiple versions anywhere because this can become a lot of work over time. For the desktop app and imports from Vikunja exports and such, we're only supporting the latest version.

I think checking the server version upfront and showing the user a screen with details about it would be a better idea. At least from a UX perspective, since it tells the user why things are not working as expected.

I see two ways to do this:

1. Blocking: do not allow the user to do anything if the server version is unsupported
2. Hint: Just tell the user "this server is not compatible and things might not work as expected" but let them continue to use the app

leschbenedikt:
I would start with version 2 to be the least blocking for the user. But I would show it on every start up. 
If that doesn't fit for the users we can change it to be blocking pretty easy or do something completely different

leschbenedikt:
I opened a new pull request for the version check #218

I think we can still merge this one additionally cause it only makes things better

kolaente (approved):

PR #215, opened 2026-01-21
fix: background refresh interval not visible in settings page

leschbenedikt:
I don't know what's happening but the diff here is completely broken for me.

It should look like that:

image


kolaente:
It seems to be fine for me when I select the option to ignore whitespace changes.

PR #218, opened 2026-01-27
feat: check for supported server version

leschbenedikt:


test/version_test.dart
```diff
@@ -0,0 +1,76 @@
...
+      Version version = Version(1, 0, 0, "beta", "1");
+      Version version2 = Version(1, 0, 0, "beta", "2");
+
+      expect(version.isNewerThan(version2), false);
+      expect(version2.isNewerThan(version), true);
+    });
+
+    test('Test rc newer', () {
+      Version version = Version(1, 0, 0, "rc1");
+      Version version2 = Version(1, 0, 0, "rc2");
```
kolaente, line 48: Can you check that `v1.0.0` is newer than `v1.0.0-rc3`?

leschbenedikt: Good point. Fixed it

Issue #211, opened 2026-01-04
[UI/UX] Immediate comment persistence conflicts with discard flow on Edit Task


RXTX4816:
**User story**
----
As a **user editing a task**, 
I expect **comment changes** to follow the same save/discard rules as other changes in that screen 
so that **leaving the screen and discarding does not result in persisted comments**.

**As is**
----
"Edit Task" screen buffers changes until "Save" is pressed and shows a discard confirmation on leave. Comment add/edit/delete are persisted immediately, so discarding does not undo them.

**To be**
----
Either buffer comment changes in Edit Task to align with Save/Discard, or move immediate comment editing to a dedicated Comments screen (a) or the task overview of a specific task (b). The discard changes confirmation on leaving the "Task Edit" screen must clearly communicate what will persist.

**Acceptance criteria**
----

- Comment changes made during Edit Task do not persist unless explicitly saved
- Leave/discard confirmation communicates which changes will be saved or discarded if there are discrepancies
- Immediate comment feature in another screen either (a) or (b) or both.

_Story points: 5_
_Reference: #190_

13rac1:
Haha! Thanks for raising this.

Buffering changes is complex and error prone; raise the story points to at least 4 for that implementation. As I wrote in [ #190](https://github.com/go-vikunja/app/pull/190#issuecomment-3661852977)

> Storing comment changes in memory will require keeping changeset history with transaction log, making multiple API calls, and probably retrying HTTP requests as needed to ensure consistency.

Moving immediate comment editing to a dedicated Comments screen is acceptable, although the use of the FAB for Save still bothers me.

RXTX4816:
I can do 5, its a valid fibonacci number 😅 With this raise I'd like to make at least the ability to instantly edit comment changes **not optional** but **required**

"Moving" might be a strong word here, I meant copying but this step was optional.

Can you elaborate what you think would be better instead of the FAB? 😆 Just so its defined. Thanks 👍

13rac1:
5, 8, or 13 😆 My point is, guaranteed consistency for a changeset of multiple object instances is complex to get right reliably. I recommend against this due to technical complexity. Example:  Task has ten comments. Delete one comment. Edit four others. Now five API requests are required on save.

As for functionality with reduced complexity, just copy the Github App, I'm using it to add this comment.

RXTX4816:
@13rac1 thanks for your thoughts on this.

[...]

PR #226, opened 2026-02-15
refactor(UI): move comments dialog to its own view

RXTX4816:
#211

### DIFF
--- a/lib/presentation/widgets/task/task_actions.dart
+++ b/lib/presentation/widgets/task/task_actions.dart
@@ -26,10 +26,8 @@ class TaskActions extends StatelessWidget {
     Navigator.push(
       context,
       MaterialPageRoute(
-        builder: (context) => TaskCommentsPage(
-          taskId: task.id,
-          taskTitle: task.title,
-        ),
+        builder: (context) =>
+            TaskCommentsPage(taskId: task.id, taskTitle: task.title),
       ),
     );
   }
@@ -57,10 +55,7 @@ class TaskActions extends StatelessWidget {
         value: _TaskAction.comments,
         child: Text(localizations.comments),
       ),
-      PopupMenuItem(
-        value: _TaskAction.edit,
-        child: Text(localizations.edit),
-      ),
+      PopupMenuItem(value: _TaskAction.edit, child: Text(localizations.edit)),
     ];
   }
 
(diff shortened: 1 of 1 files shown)


leschbenedikt:
I don't think we should put a button in the list item itself since it uses a lot of space that could be used to display the title.

Also we could restore the old save FAB as we do not interfere with the comments anymore

RXTX4816:
@leschbenedikt thanks for the review

1) Was thinking the same. But then lets please also remove the "Edit Button" from the list view, since its also taking some space away. It can be solved better:
Proposals: 
a) Have a 3 dot button (Burger button) to show all actions of a task in the list view
b) Have no buttons in List view, but long press on a task to access its actions.

2) Yepp good idea, will do it :)

RXTX4816:
@leschbenedikt ready for review.

Changes:
- ~~background_downloader version mismatch -> Fixes pipeline~~
- ~~Test and UI had some wrong variable name -> Fixes pipeline~~
- Added context menu actions (Long press on task or the 3 dot button)
- The FAB Save button was already there, no change here needed

leschbenedikt:
If I open it on one of my emulator it looks like this
image


Also why do we need the long press context menu?

RXTX4816:
@leschbenedikt 

On my emulator it also looks like that. I had a similar case a few months ago with another PR here, and on an actual phone it worked fine so I did not put any attention to it. I did not test this PR on an actual phone but will ping you once I find time to do so. Edit: I think you are right, there might be a problem with the placement.

We certainly don't need the long press context menu. I don't even know if its standard for current Android, it was a lot of years ago for classic Android. I went through other apps and some implement it like that, some don't. I personally don't really care about that. The [material docu](https://m2.material.io/components/menus/android#dropdown-menus) acknowledges such long-press context menu can exist, so its not entirely wrong - we just need to decide if it fits this app 👍 
My decision to put the long press context menu was because currently single press and long press already have the same function - to open the overview of the pressed task - so I was thinking to have (in the future) a way to press, hold (context opens) and then drag your finger to the desired context menu action and select the action on a depress.

RXTX4816:
Ready for review @leschbenedikt . Changes:

- fixed overflow icon
- removed long-press context menu
- created new widget on task actions to minimize duplicated code for the context menu with two flavors:
  - menu for the burger bar action with strings
  - icon for the task bottom sheet (task overview) with icons only.

I decided to remove the long-press context menu since it does not really belong to this ticket. I may create another issue for it to discuss whether that functionality is desirable.

PR #228, opened 2026-02-17
Several fixes 

### LINKED ISSUE #223: Bug: Task list does not update after deleting a task from the edit screen
If you open a task from the list, then delete it via the delete button in the edit screen, the edit screen closes but the task list is not updated. The deleted task still appears until the list is refreshed manually 

**Steps to reproduce:**

1. Open a project/list that shows a list of tasks.
2. Tap a task to open the task edit screen.
3. Tap the delete (trash) icon in the app bar.
4. Confirm deletion in the dialog.
5. The edit screen closes and you return to the task list.

leschbenedikt:
- Resolves #223 
- Supported server version update to 1.1.0
- Nullpointer in settingscontroller because it was initialized before login
- Nullpointer when created by not set in task
- Type mismatch for project view position fixed
- Dependency update
- Cleaned up android build file

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -165,10 +165,10 @@ packages:
     dependency: transitive
     description:
       name: characters
-      sha256: faf38497bda5ead2a8c7615f4f7939df04333478bf32e4173fcb06d428b5716b
+      sha256: f71061c654a3380576a52b451dd5532377954cf9dbd272a78fc8479606670803
       url: "https://pub.dev"
     source: hosted
-    version: "1.4.1"
+    version: "1.4.0"
   checked_yaml:
     dependency: transitive
     description:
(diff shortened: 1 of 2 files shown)


kolaente:
Seems like the CI is failing now:

```
Failed to parse /home/runner/work/app/app/.dart_tool/package_graph.json: dependencies for `flutter_inappwebview_windows` missing.
```

Since we don't support Windows here, I don't think we'll need that dependency at all? Not sure if we can control that though

leschbenedikt:
The error message was misleading.
The issue occurred because of a manual overwrite when do not need anymore.

But also although we don't support it be do have build targets for windows, linux and macos.
Think we should fix that in the future

PR #229, opened 2026-02-17
fix: type cast failure

leschbenedikt:


kolaente:
Would it make sense to create test cases around these type failures?

leschbenedikt:
Yes. Is on my agenda. I would like to write a lot more test cases to be more resilient to regression

PR #232, opened 2026-02-17
Remaining code todos

leschbenedikt:


lib/core/utils/repeat_after_parse.dart
```diff
@@ -1,39 +1,50 @@
...
   if (repeatAfter == null || repeatAfter.inSeconds == 0) {
-    return null;
+    return 0;
   }
 
   // if its dividable by 24, its something with days, otherwise hours
   if (repeatAfter.inHours % 24 == 0) {
     if (repeatAfter.inDays % 365 == 0) {
-      return 'Years';
+      return 4; //Years
```
kolaente: Maybe worth doing this with some kind of custom type or at least consts? That would make it explicit, not relying on numbers

kolaente (approved):

lib/core/utils/repeat_after_parse.dart
leschbenedikt: Changed it. Looks way better now. Thanks

PR #237, opened 2026-02-27
feature: adapt app to new login flow

leschbenedikt:
Resolves https://github.com/go-vikunja/app/issues/238

### DIFF
--- a/lib/core/network/client.dart
+++ b/lib/core/network/client.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert';
 import 'dart:core';
 import 'dart:developer' as developer;
@@ -10,6 +11,7 @@ import 'package:http/io_client.dart' as io_client;
 import 'package:logging/logging.dart';
 import 'package:sentry_flutter/sentry_flutter.dart';
 import 'package:vikunja_app/core/network/response.dart';
+import 'package:vikunja_app/data/data_sources/settings_data_source.dart';
 import 'package:vikunja_app/main.dart';
 import 'package:vikunja_app/presentation/widgets/string_extension.dart';
 
@@ -21,7 +23,12 @@ class Client {
 
   String _token = '';
   String _base = '';
+  String? refreshCookie;
   bool ignoreCertificates = false;
+
+  SettingsDatasource? settingsDatasource;
+
+  Completer<bool>? _refreshCompleter;
 
   late http.Client _httpClient;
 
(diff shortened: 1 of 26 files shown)


lib/core/network/client.dart
```diff
@@ -198,6 +263,98 @@ class Client {
...
+    if (_refreshCompleter != null) {
+      return _refreshCompleter!.future;
+    }
+
+    _refreshCompleter = Completer<bool>();
+
+    try {
+      final refreshHeaders = {
+        'Content-Type': 'application/json',
+        'User-Agent': 'Vikunja Mobile App',
```
kolaente: We should centralize this - ideally it would include the version as well and be the same string for all requests the app makes.

lib/core/network/client.dart
```diff
@@ -198,6 +263,98 @@ class Client {
...
+  }
+
+  Future<Response<T>> _handleResponseWithRefresh<T>(
+    http.Response response,
+    T Function(dynamic body)? mapper,
+    Future<http.Response> Function() retryRequest,
+  ) async {
+    _extractRefreshCookie(response.headers);
+
+    if (response.statusCode == 401 && _refreshCookie != null) {
```
kolaente: This should also check the error code: https://vikunja.io/docs/api-login-session-migration/#error-response-change

lib/core/utils/constants.dart
```diff
@@ -1,7 +1,7 @@
 import 'package:flutter/material.dart';
 import 'package:vikunja_app/domain/entities/version.dart';
 
-var supportedServerVersion = Version(1, 1, 0);
+var supportedServerVersion = Version(2, 0, 0);
```
kolaente: Since we're now at 2.1.0, let's increase that as well.

lib/data/data_sources/settings_data_source.dart
```diff
@@ -148,7 +148,14 @@ class SettingsDatasource {
     return _storage.write(key: "user-token", value: token);
   }
 
-  // Locale override (if null or empty -> use system locale)
+  Future<String?> getRefreshCookie() {
+    return _storage.read(key: "refresh-cookie");
+  }
+
+  Future<void> saveRefreshCookie(String? cookie) {
+    return _storage.write(key: "refresh-cookie", value: cookie);
```
kolaente, line 156: (this might be a whole new can of worms)

Is this storing the string in some kind of secret storage? I'm not exaclty sure on how this works, but IIRC there are ways to store data in ways that other apps cannot access them.

leschbenedikt: Yes, it's in a secure storage. It is using FlutterSecureStorage which uses encrypted storages on android an ios

PR #244, opened 2026-03-05
Add button on widget to add a task quickly

brokoli18:
Currently this button opens the main app and launches the add-task popup. The only downside right now is if the app is closed and has to be opened from the widget it can take a few seconds. Ideally a small/custom activity can be created and launched but that is more complex and for another day.


brokoli18:
@leschbenedikt @kolaente I think this one is ready for review

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -149,9 +176,18 @@ class AppWidget : GlanceAppWidget() {
...
+            TitleBar(
+                title = "Vikunja",
+                startIcon = ImageProvider(R.drawable.vikunja_logo),
+                iconColor = null,
+                actions = {
+                    CircleIconButton(
+                        enabled = true,
+                        onClick = actionRunCallback<InteractiveAction>(),
+                        imageProvider = ImageProvider(R.drawable.add),
+                        contentDescription = "Add a Task",
```
RXTX4816: If an User that has accessibility enabled (e.g. Vision-impaired with Talkback on) and does not understand english it could lead to frustration. 
Unfortunately the Widget Strings cannot read the Flutter Translations. 
Since this is the first time we use language-dependend strings in the widgets there are 2 options:
- Add pipeline step to create strings.xml from .arb files (Promising)
- Hard-code strings.xml for each language (Meh)

I think its fine for now with hard-coded english string (as in not blocking), just wanted to comment this so for the future we take a look.

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -42,7 +44,32 @@ import java.time.format.DateTimeFormatter
...
 
+class InteractiveAction : ActionCallback {
+    override suspend fun onAction(
+        context: Context,
+        glanceId: GlanceId,
+        parameters: ActionParameters
+    ) {
+
+        val intent = Intent(context, MainActivity::class.java).apply {
+            putExtra("openAddTask", true)
```
RXTX4816: Do we need this line? I can't find any mention of "openAddTask" in the app. If I remove this line it still works fine, if I see correctly the app takes action and type.

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
```diff
@@ -149,9 +176,18 @@ class AppWidget : GlanceAppWidget() {
...
         ) {
-            Text(
-                text = "Today",
-                style = TextStyle(fontSize = 20.sp, color = ColorProvider(Color.White, Color.White))
+            TitleBar(
+                title = "Vikunja",
+                startIcon = ImageProvider(R.drawable.vikunja_logo),
+                iconColor = null,
+                actions = {
+                    CircleIconButton(
```
RXTX4816: On my emulator the circle is cut-off horizontally at the top and bottom part.
This is probably due to low resolution, but there are Users who use phones with low resolution. 
Possible fixes:
- increasing the bar height to 56.dp or
- removing the fixed height and letting TitleBar size itself 

Attached screenshot:

Image

RXTX4816 (commented):
Good work I really enjoy this feature. I tested it locally and it works great! Added some small comments from my side after reviewing code but no biggies.

android/app/src/main/kotlin/io/vikunja/flutteringvikunja/widget/AppWidget.kt
brokoli18: @RXTX4816 I put the icon in a box and added some padding so that its no longer cut off. Have a look and let me know if it works

RXTX4816: Hey just tested it looks good now on my emulator 👍

brokoli18:
I think everything is sorted here and this is ready to merge

leschbenedikt (approved):

PR #248, opened 2026-03-10
fix: logout after concurrent token update

### LINKED ISSUE #246: Frequent app logouts
Not sure if similar reason to https://github.com/go-vikunja/app/issues/245 but since I don't use OIDC decided to make a new issue.

When background refresh is enabled, the app does not stay logged in, and prompts re-login each time I open it up.

Looking at my server logs, it looks like the app requests token refresh, then immediately tries to refresh with the old token, which causes it to fall back to entering credentials.

Logs from my server (specifically lines 3 and 4 seem suspect

```
VikunjaServer[1183]: time=2026-03-09T13:54:20.022Z level=INFO component=http method=GET uri="/api/v1/tasks?filter=done+%3D+false+%26%26+due_date+%3C+now%2Fd%2B1d" status=401 latency=5.001898ms
VikunjaServer[1183]: time=2026-03-09T13:54:20.022Z level=INFO component=http method=GET uri="/api/v1/tasks?filter=done%3Dfalse+%26%26+%28due_date+%3E+now+%7C%7C+reminders+%3E+now%29&filter_include_nulls=false" status=401 latency=3.317754ms
VikunjaServer[1183]: time=2026-03-09T13:54:20.195Z level=INFO component=http method=POST uri=/api/v1/user/token/refresh status=200 latency=54.581557ms
VikunjaServer[1183]: time=2026-03-09T13:54:20.197Z level=ERROR component=http method=POST uri=/api/v1/user/token/refresh status=401 latency=57.021137ms err="code=401, message=Refresh token already used."
VikunjaServer[1183]: time=2026-03-09T13:54:20.277Z level=INFO component=http method=GET uri="/api/v1/tasks?filter=done+%3D+false+%26%26+due_date+%3C+now%2Fd%2B1d" status=200 latency=29.519731ms
VikunjaServer[1183]: time=2026-03-09T13:54:24.504Z level=INFO component=http method=GET uri=/api/v1/projects status=401 latency=129.461µs
VikunjaServer[1183]: time=2026-03-09T13:54:24.635Z level=ERROR component=http method=POST uri=/api/v1/user/token/refresh status=401 latency=1.735218ms err="code=401, message=Invalid or expired refresh token."
VikunjaServer[1183]: time=2026-03-09T13:54:24.695Z level=INFO component=http method=GET uri=/api/v1/projects status=401 latency=138.739µs
VikunjaServer[1183]: time=2026-03-09T13:54:24.813Z level=ERROR component=http method=POST uri=/api/v1/user/token/refresh status=401 latency=1.660071ms err="code=401, message=Invalid or expired refresh token."
```

EDIT:
Looking through code, it seems that the lib/core/network/client.dart::Client updates the refresh cookie in global store, but never actually reads it after it was initialized. Could there be a race somewhere with multiple Client instances refreshing token at the same time?
Happy to try my hand at handling this

leschbenedikt:
Fixes #246


lib/core/network/token_lock.dart
```diff
@@ -0,0 +1,43 @@
...
+        await file.create();
+      } catch (_) {
+        // Ignore likely race condition on creation
+      }
+    }
+
+    final raf = await file.open(mode: FileMode.append);
+
+    try {
+      await raf.lock(FileLock.blockingExclusive);
```
msdinit: Looking at docs
https://api.flutter.dev/flutter/dart-io/RandomAccessFile/lock.html

> On Linux and OS X this uses advisory locks, which have the surprising semantics that all locks associated with a given file are removed when any file descriptor for that file is closed by the process. Note that this does not actually lock the file for access. Also note that advisory locks are on a process level. This means that several isolates in the same process can obtain an exclusive lock on the same file.

does not look like it would work. Tried on a minimal app on emulated android and it looks like isolates can concurrently lock the file

msdinit: Perhaps you can treat the fact that file exists as a lock?
Additionally if the file exists we can check `lastAccessed` or `lastModified` and discard the lock if it was creation time is older than some constant in case isolate that created the lock was killed

leschbenedikt: We don't actually support Linux or OS X although the project suggests it.
And Android normally counts as separate platform to Linux in Flutter.

What is more worrying is that you could do it on android.
Could you provide you example code?

msdinit: Sure, here's the repo:
https://github.com/msdinit/dart_lang_test

Most of the code is just scaffolding to deploy on emulator, the test itself is in main:
https://github.com/msdinit/dart_lang_test/blob/master/lib/main.dart#L15

And here are the logs, showing that both isolates are holding the lock at the same time:
```
[log] Iso 1: file exists
[log] Iso 1: Successfully opened
[log] Iso 1: Start sleeping
[log] Iso 2: file exists
[log] Iso 2: Successfully opened
[log] Iso 2: Start sleeping
[log] Iso 1: Unlocked
[log] Iso 2: Unlocked
```

leschbenedikt: This was way more complicated than I thought.
I rewrote the locking logic and changed to test case so it actually runs on an emulator.
It worked for me in a short test. Will test more

msdinit: Thank you for your hard work!
I tested this branch a bit, the token refresh seems to be fixed.
It looks like background task refresh task is succeeding, but not actually getting the newly created tasks or scheduling new notifications, but I'll test it a bit more later today, could be an issue with my setup

leschbenedikt: Good catch. I already found an issue that I fixed.
But somehow the background task just stops.

I added a lot of debug log and it just stops on random spots during execution without visible error.
I try to investigate further but as for now I am a little confused 😅

leschbenedikt: Found the issue.
Some awaits were missing so the isolate stopped before actually done.

Should work now

msdinit: Works great!
Been running on emulator for several hours, getting new notifications scheduled in background successfully and no logouts whatsoever

.vscode/launch.json
```diff
@@ -6,14 +6,22 @@
...
     {
       "name": "Flutter",
       "request": "launch",
       "type": "dart",
       "flutterMode": "debug",
-      "args": ["--flavor", "unsigend"]
+      "args": ["--flavor", "unsigned"]
+    },
+    {
+      "name": "Integration Test: Concurrent Refresh",
```
kolaente: Please leave that out of the launch file, but the test should be run in CI

lib/core/network/client.dart
```diff
@@ -265,87 +243,64 @@ class Client {
...
-      if (_refreshCompleter != null && !_refreshCompleter!.isCompleted) {
-        _refreshCompleter!.complete(false);
-      }
-      _refreshCompleter = null;
       return false;
     }
   }
 
-  void _extractRefreshCookie(Map<String, String> responseHeaders) {
+  String? _extractRefreshCookie(Map<String, String> responseHeaders) {
```
kolaente: This seems duplicated with the function of the same name in the user data source, can we extract them to a shared helper?

kolaente: Similarly, we're checking the refresh token cookie on every request, but there are only three routes that actually return a refresh cookie. Can we simplify this?

leschbenedikt: Yeah right seems a little overkill.
And on login we do it twice.

I know /login and /refresh can return one. What's the third route?

leschbenedikt: Done

kolaente:
Please resolve the conflicts

kolaente (approved):

CrazyWolf13:
@kolaente is this live yet in the playstore version?

Is there any possibility on getting releases or at least tags here on github as well?

leschbenedikt:
Yes this is live in the play store beta.

We are on the road to 1.0 but there a few things left to do

PR #250, opened 2026-03-23
feat: enable background refresh by default (60 min)

### LINKED ISSUE #185: Background Refresh disabled by default
This being disabled will mean that the app will never refresh in the background unless the user goes in and explicitly changes this value in the settings. Although it isn't an outright bug i would think that most users would want this kind of application to update in the background by default

kolaente:
Background refresh was disabled by default for new installs, meaning users had to manually discover and enable it. This changes the default to 60 minutes, re-registers the periodic task on every app startup, and fixes iOS where the missing `BGTaskSchedulerPermittedIdentifiers` key caused `workmanager` periodic tasks to silently fail.

- Default refresh interval changed from 0 to 60 minutes (existing user settings preserved)
- WorkManager periodic task re-registered at every app launch
- Extracted shared `registerBackgroundRefresh()` to eliminate duplication between `main()` and `SettingsController`
- Added `BGTaskSchedulerPermittedIdentifiers` to iOS `Info.plist` (required since iOS 13)

Resolves https://github.com/go-vikunja/app/issues/185

### DIFF
--- a/lib/core/background_work.dart
+++ b/lib/core/background_work.dart
@@ -71,3 +71,22 @@ Future<bool> updateTasks() async {
 
   return Future.value(true);
 }
+
+/// Registers (or cancels) the periodic background refresh task with the OS.
+///
+/// Call this at app startup and whenever the user changes the interval.
+/// If [minutes] is 0, all existing tasks are cancelled and nothing is registered.
+Future<void> registerBackgroundRefresh(int minutes) async {
+  if (kIsWeb) return;
+
+  await Workmanager().cancelAll();
+  if (minutes > 0) {
+    await Workmanager().registerPeriodicTask(
+      "update-tasks",
+      "update-tasks",
+      frequency: Duration(minutes: minutes),
+      constraints: Constraints(networkType: NetworkType.connected),
+      initialDelay: Duration(seconds: 15),
+    );
+  }
+}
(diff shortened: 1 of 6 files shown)


lib/main.dart
```diff
@@ -74,6 +74,8 @@ void main() async {
   try {
     if (!kIsWeb) {
       Workmanager().initialize(callbackDispatcher);
+      final refreshInterval = await settingsDatasource.getRefreshInterval();
```
leschbenedikt, line 77: I don't really know about this but it somehow feels wrong to do this on every startup.
Maybe we should introduce something like a "isFirstStartup"

lib/core/background_work.dart
```diff
@@ -77,3 +77,22 @@ Future<bool> updateTasks() async {
...
 }
+
+/// Registers (or cancels) the periodic background refresh task with the OS.
+///
+/// Call this at app startup and whenever the user changes the interval.
+/// If [minutes] is 0, all existing tasks are cancelled and nothing is registered.
+Future<void> registerBackgroundRefresh(int minutes) async {
+  if (kIsWeb) return;
+
+  await Workmanager().cancelAll();
```
leschbenedikt, line 82: I think we should cancel by name and not all.
If we ever use workmanager for something else we will cancel that task too

Issue #258, opened 2026-04-16
[Bug] Init screen forever loading on app start if the server does not respond


RXTX4816:
User Story
----
As a user I want to be able to make changes to the configuration about which server to connect to when my currently connected server experiences problems.

As is
----
When the vikunja server does not produce a response and the app was previously connected, on an app start, the app shows the "Loading..." animation indefinetly without any error message.

To be
----
When the vikunja server does not produce a response within a defined timeout, make it possible for the user to go back to the screen that lets them select a server or try again.

Steps to reproduce
----
1) Start server, start app, login to server.
2) docker pause vikunja (Pause the server or basically make it not produce a response)
3) Kill the app and start the app.

Screenshot:

<img width="874" height="1560" alt="Image" src="https://github.com/user-attachments/assets/3755f360-859f-4e10-ba27-46416f2b3ad7" />

leschbenedikt:
Resolved with #259

PR #259, opened 2026-04-16
fix(client): forever loading when no server response now timeouts with error handling

RXTX4816:
#258

If the server does not produce response when starting the app, let the user be able to logout (=return to server configuration screen)

image


lib/init_page.dart
```diff
@@ -6,66 +6,90 @@ import 'package:vikunja_app/main.dart';
...
-                  },
-                );
-              }
-              globalNavigatorKey.currentState?.pushReplacementNamed('/home');
-          }
-        });
-      });
-    });
+class _InitPageState extends ConsumerState<InitPage> {
+  bool _handledOutcome = false;
```
leschbenedikt: Why do we need this flag?

leschbenedikt (commented):
Looks motly good. Just the flag that I don't really understand at the moment

lib/init_page.dart
RXTX4816: Thanks for reviewing!

The flag prevents the outcome from being handled multiple times if the widget rebuilds while in the same state. Without it there could be duplicate navigation or dialogs.

Update: I implemented it with riverpod now - works also perfectly fine. Or you meant we dont need that flag at all?

leschbenedikt: Was just wondering why this can happen. But looks even better now.
This build is broken for now. #275 will fix that.
After that we can merge

leschbenedikt (approved):

PR #266, opened 2026-05-12
Allow hierarchical view of task and subtasks

simson:
I use heavily the subtask feature. I want them to be presented hierarchically.
A settings is available to set this feature on/off. It's disable by default to keep the current behaviour.

This PR is done by human, but the code was written by AI. If i miss a mention of NO-AI allowed, please excuse me.

### DIFF
--- a/lib/data/models/task_dto.dart
+++ b/lib/data/models/task_dto.dart
@@ -66,7 +66,12 @@ class TaskDto extends Dto<Task> {
       dueDate = DateTime.parse(json['due_date']),
       startDate = DateTime.parse(json['start_date']),
       endDate = DateTime.parse(json['end_date']),
-      parentTaskId = json['parent_task_id'],
+      parentTaskId = json['parent_task_id'] ??
+          (json['related_tasks'] != null &&
+                  json['related_tasks']['parenttask'] != null &&
+                  (json['related_tasks']['parenttask'] as List).isNotEmpty
+              ? (json['related_tasks']['parenttask'] as List)[0]['id'] as int
+              : null),
       priority = json['priority'],
       repeatAfter = Duration(seconds: json['repeat_after']),
       color = json['hex_color'] != ''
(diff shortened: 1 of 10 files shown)


leschbenedikt:
Hey,
thanks for the contribution.
As far as I know there is no ban of AI code.
But I hardly recommend checking and understanding the produced code.

I will try to review the code within the next week.
For now, to fix the build you have to format the code correctly.

Just run `dart format .` and it will be formatted correctly

lib/presentation/pages/settings_page.dart
```diff
@@ -121,6 +122,21 @@ class SettingsPageState extends ConsumerState<SettingsPage> {
                 },
               ),
               Divider(),
+              SwitchListTile(
```
leschbenedikt, line 125: I don't think we need this to be configurable
Is this configurable in the web page?
If not I would suggest to remove it

lib/data/models/task_dto.dart
```diff
@@ -66,7 +66,12 @@ class TaskDto extends Dto<Task> {
       dueDate = DateTime.parse(json['due_date']),
       startDate = DateTime.parse(json['start_date']),
       endDate = DateTime.parse(json['end_date']),
-      parentTaskId = json['parent_task_id'],
+      parentTaskId = json['parent_task_id'] ??
```
leschbenedikt, line 69: Did the api change here or was it always wrong? Just wondering.
Anyway, we agreed on just supporting one version of the api for now to reduce complexity and maintenance. 
So you can remove the `json['parent_task_id']` part

Also as a task can have multiple parents this should be a list

lib/data/models/task_dto.dart
leschbenedikt, line 91: Same as above. The `json['subtasks']` part should be removed

lib/presentation/pages/task/task_list_page.dart
```diff
@@ -68,28 +74,111 @@ class TaskListPage extends ConsumerWidget {
...
+  }
+
+  Widget _buildHierarchicalList(
+    WidgetRef ref,
+    BuildContext context,
+    TaskPageModel model,
+  ) {
+    // Build subtask map from each parent's .subtasks list so that a task
+    // with multiple parents appears under all of them.
+    final taskById = {for (final t in model.tasks) t.id: t};
```
leschbenedikt, line 125: Why do I need to do this.
`task.subtasks` is basically the list you a creating

lib/presentation/pages/task/task_list_page.dart
```diff
@@ -68,28 +74,111 @@ class TaskListPage extends ConsumerWidget {
...
+      if (task.subtasks.isNotEmpty) {
+        subtaskMap[task.id] =
+            task.subtasks.map((s) => taskById[s.id] ?? s).toList();
+        for (final s in task.subtasks) {
+          subtaskIds.add(s.id);
+        }
+      }
+    }
+
+    final topLevelTasks =
```
leschbenedikt, line 138: Isn't this just `where((t) => t.subtasks.isEmpty())` ?

leschbenedikt (commented):
I didn't review all of it just a few things I noticed right away.

The biggest problem is that it didn't work for me at all.
The only thing that changed for me is that the list items are now looking a little strange

Image

Only the bottom one has a sub task which is not visible

PR #267, opened 2026-05-12
feat: add Linux desktop support

simson:
Add Linux platform target (flutter create --platforms=linux).

Update the GTK runner to support single-instance mode so that deep links (vikunja-flutter://callback) are forwarded to the running instance via the GApplication command-line signal:
- Replace G_APPLICATION_NON_UNIQUE with G_APPLICATION_HANDLES_COMMAND_LINE | G_APPLICATION_HANDLES_OPEN
- Return FALSE from local_command_line so GApplication processes the command line (required for the gtk plugin to receive it)
- Present existing window instead of creating a duplicate on activate

Guard mobile-only plugin calls in main.dart behind platform checks:
- permission_handler (no Linux implementation)
- workmanager (no Linux implementation, was also missing await so the UnimplementedError escaped the try/catch)

To register the OAuth deep-link URL scheme on the host, run once:
  cp <bundle>/vikunja_app.desktop ~/.local/share/applications/
  xdg-mime default io.vikunja.vikunja_app.desktop x-scheme-handler/vikunja-flutter
  update-desktop-database ~/.local/share/applications/

Build command:
  JAVA_HOME=<jdk17> LIBRARY_PATH=/usr/lib/gcc/x86_64-linux-gnu/11 \
  flutter build linux --debug

### DIFF
--- a/.metadata
+++ b/.metadata
@@ -4,7 +4,7 @@
 # This file should be version controlled and should not be manually edited.
 
 version:
-  revision: "300451adae589accbece3490f4396f10bdf15e6e"
+  revision: "00b0c91f06209d9e4a41f71b7a512d6eb3b9c694"
   channel: "stable"
 
 project_type: app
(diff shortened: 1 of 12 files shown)


RXTX4816:
Hi I have a few questions/comments:

1) What are the intention and benefits behind adding a linux build here to the mobile app, in contrast to using the [Desktop app](https://dl.vikunja.io/desktop/v2.3.0/), creating a seperate native Desktop app or using an Android Emulator/APK runner under linux? 

2) Are all Linux desktop envs and window managers supported? (E.g. KDE/QT env's without GTK, tiling WMs like i3, other WM/DE that use wayland(Looking at proper scaling and performance))

3) How would this be finally installed from an end-users perspective?

simson:
1) my purpose was just to ease the development as android emulator was complicated to setup for me. 
2) i don't even try. it was just to devellop the hierchical feature #266 
3) i don't know : i won't use it myself. 

so to conclude, i don't think this is pertinent. i thought it was maybe usefull, you're questions showned me that it is not. thanks for your time

RXTX4816:
No worries :)

Yeah setting up an emulator can be tough, I had good experience with the following chain on linux/macos:

sdkmanager, avdmanager, emulator

Take care!

PR #268, opened 2026-05-18
fix: use correct action id for due date notification

msdinit:
I think this typo is what caused due date notifications 'Done' button to do nothing


msdinit:
Closing in favor of #270

PR #270, opened 2026-05-24
Android reminder and due date notifications silently fail on Android 12+

### LINKED ISSUE #98: push notifications stopped working
Hi, when I add a task with a due date and a reminder, the app does not show any notification when the task is due.
Just a couple of days ago it was working fine, on the same vikunja instance v. 0.24.4
I already tried reinstalling the app (v. 0.1.8). The test notification is working fine

Samsung Galaxy s24 ultra, android 14, october security patch

adejong5:
## Problem

On Android 12+ (API 31+), `SCHEDULE_EXACT_ALARM` is not granted by default and must be requested at runtime. The app calls `zonedSchedule()` with `exactAllowWhileIdle` but never checks or requests this permission, causing all scheduled reminder and due date notifications to fail silently. Test notifications still work because they use `show()` (immediate delivery) rather than `zonedSchedule()`.

Additionally, a typo introduced in a3777dcf sets the due date notification action id to `'action_dcd one'` instead of `'action_done'`, so tapping Done on a due date notification does nothing.

Closes #98

## Changes

- Call `requestExactAlarmsPermission()` during `initNotifications()`, guarded by `canScheduleExactNotifications()` so devices that already have the permission (e.g. via `USE_EXACT_ALARM`) are unaffected
- Fix the `'action_dcd one'` typo by extracting the action id into a named constant `_notificationActionDone` used across all three action definitions and the background handler. This supersedes #268, which fixes the typo alone
 
## Testing
Tested on a Pixel 10 (Android 16, API 36) against a self-hosted Vikunja instance:
- Reminder notification fires while app is in foreground ✓
- Reminder notification fires after app is swiped from recents ✓
- Done action on both reminder and due date notifications marks task complete ✓
- `flutter test` — 174 tests pass ✓
- `dart format --set-exit-if-changed` — no changes ✓

### DIFF
--- a/lib/presentation/manager/notifications.dart
+++ b/lib/presentation/manager/notifications.dart
@@ -15,12 +15,13 @@ import 'package:vikunja_app/domain/repositories/task_repository.dart';
 import 'package:vikunja_app/presentation/manager/widget_controller.dart';
 
 const _actionDonePortName = 'action_done_port_name';
+const _notificationActionDone = 'action_done';
 
 @pragma('vm:entry-point')
 Future<void> notificationTapBackground(
   NotificationResponse notificationResponse,
 ) async {
-  if (notificationResponse.actionId == "action_done") {
+  if (notificationResponse.actionId == _notificationActionDone) {
     var id = notificationResponse.id;
 
     if (id != null) {
@@ -78,7 +79,7 @@ class NotificationHandler {
     icon: 'vikunja_notification_logo',
     importance: Importance.high,
     actions: <AndroidNotificationAction>[
-      AndroidNotificationAction('action_dcd one', 'Done'),
+      AndroidNotificationAction(_notificationActionDone, 'Done'),
     ],
   );
   var androidSpecificsReminders = AndroidNotificationDetails(
@@ -88,7 +89,7 @@ class NotificationHandler {
     icon: 'vikunja_notification_logo',
     importance: Importance.high,
     actions: <AndroidNotificationAction>[
-      AndroidNotificationAction('action_done', 'Done'),
+      AndroidNotificationAction(_notificationActionDone, 'Done'),
     ],
   );
   late DarwinNotificationDetails iOSSpecifics;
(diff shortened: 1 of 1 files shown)


lib/presentation/manager/notifications.dart
```diff
@@ -114,6 +115,20 @@ class NotificationHandler {
     initBackgroundCommunication();
 
     requestIOSPermissions();
+    await _requestAndroidExactAlarmPermission();
+  }
+
+  Future<void> _requestAndroidExactAlarmPermission() async {
```
leschbenedikt, line 121: I'm not sure about this.

First of all because notification work me. So not all notifications fail silently.
And if at least some work, it can't be a permission issue in my opinion.

Also neither the documentation of the flutter library https://pub.dev/packages/flutter_local_notifications
nor the official android documentation https://developer.android.com/develop/ui/compose/notifications/notification-permission 
mention this

If for some reason notifications do not work on your end I'm glad to look into this but I don't think this is the reason.

What version of the app did you use and what phone?

adejong5: The methods used to set alarms (`setExactAndAllowWhileIdle`) of the `AlarmManager` class [require the `SCHEDULE_EXACT_ALARM` permission](https://developer.android.com/reference/android/app/AlarmManager#setExactAndAllowWhileIdle(int,%20long,%20android.app.PendingIntent)). 

image

While that is set in the manifest, the permission [must be granted using `requestExactAlarmPermissions`](https://pub.dev/packages/flutter_local_notifications#androidmanifestxml-setup) 

image

I tested on Pixel 10 Android 16. Due date notifications failed before, as explained in the 2-year-old open Issue #98. Forked, implemented this, tested using an flutter debug install and an unsigned release and it works. Permissions are inherited during Android version updates so it's possible you have `SCHEDULE_EXACT_ALARM` permissions grandfathered in from when they were approved automatically?

leschbenedikt: Could be related to #273 

After looking deeper into that, it seems you're right and we need this permission.

But what I don't get is why it does work for me. I tested it on my old S24 Ultra and on the S26 Ultra. Even tested it on a Pixel 10 in the Emulator. Worked every time.

Also I don't think I ever updated the android version since I use the app especially on the S26.

adejong5: Strange. I think I saw a note about debug versions being giving permissions automatically, but I can't find where I saw that again. There was also a note about not needing `SCHEDULE_EXACT_ALARM` if battery protections were disabled.Honestly, I don't entirely understand the difference between `SCHEDULE_EXACT_ALARM` and `USE_EXACT_ALARM` but I can't argue with the results.

leschbenedikt: That could explain why it works on the Emulator but on my phone I use the release version.
I will investigate that further because I want to understand whats happening.
But cause we need that permission anyway let's merge

leschbenedikt (approved):

Issue #257, opened 2026-04-12
Material You icon is muted.


martijnw1986:
When "themed" Material You icons are enabled on Android, Vikunja is the only one with a muted look, as if the app is disabled. 

![Image](https://github.com/user-attachments/assets/73a0140c-8419-4ff6-916b-1423436114bc)

leschbenedikt:
Hi, 
How did you theme it to look like this. Or what device, android version and launcher are you using.
At the moment I don't think this is a bug but a result on how the theming work on you're end.

But I will try look deeper into this once I know how to reproduce it

martijnw1986:
Hi [leschbenedikt](https://github.com/leschbenedikt),

I'm using a Pixel 10 running Android 16. I'm using no third-party apps to theme my device. Using the default Pixel Launcher, when you set a wallpaper, it opens Google's Wallpaper app. When you set a wallpaper, it creates a palette of colours based on that wallpaper. In the same wallpaper app, you can set the icons to match the colour-palette. Most modern apps support this and the few that don't display the standard full-colour icon. Vikunja is the first app that I have encountered to support the feature but render the background colour in a muted tone.

Thank you for your response. Let me know if there is anything else that I can do.

leschbenedikt:
The good thing is that I could reproduce it.
But we never intended to support this and I still don't know why the phone is trying to do so 😆 

I just played around and it looked as intended. But to make this work I need to really design the icon to look right.
And I'm not much of a designer
It doesn't have I high priority right now but I will try to support this sometime in the near future.

skiwichu:
Hmm, I would appreaciate it too 🥲

PR #284, opened 2026-07-01
feature: support for monochrome icons

leschbenedikt:
#257


android/app/src/main/res/mipmap-xxhdpi/ic_launcher.png
kolaente, line 1: Why does the new version have the slight curve on the neck? The previous didn't have it

leschbenedikt: You're right. Didn't notice it at all. That must have come from the icon generator.
Reverted the ones in question.

Issue #206, opened 2025-12-20
Widget customization


scoop96:
Hello, first thank you so much for your hard work and helping my life to be easier.

Regarding the widget, its very poor, i would love to be able to set which projects to be shown (and any range of date (or all the to-do's))

I dont usually put end date (because i organize myself in projects) and the current widget only shows "today" and my widget is always empty.

Thank you

leschbenedikt:
Thanks for the request.
I like the idea and would love to see that.
But I can't promise this will come anytime soon as there a lot of other things to do in the app that
mostly have a higher priority.

But feel free to open a PR

ilvez:
I also have some adjustment ideas (font size, spacing) for the widget. 

Glad you are open for PRs. I already did some local changes that solved it for me, but now after 1.1.0 got released and already the app has moved forward, I don't want to live with backporting those features locally any longer.

So lets see if I can throw out some PRs for widget.

PR #287, opened 2026-07-04
Make widget customizable by project or upcoming views

bgrassy:
This is a pass at addressing #206. The initial layers of customization exposed are filtering projects by upcoming, today, or by project. I did not do much customization on the UI of the settings yet, just currently complying with system light/dark mode.

I also fixed a visual bug where the widget background would be clear if there weren't enough tasks to fill the whole size of the widget.

I added some logic to account for multiple widgets, and tweaked the date representation to account for some serialization errors I was hitting. I wrote the bulk of the logic myself, but used Claude to mock initial review and fix issues caught there. I tested on my own device and it has worked, but I am not a power user yet so recommendations on further testing/edge cases would be appreciated!

Screenshot_20260704-085007
Screenshot_20260704-084958
Screenshot_20260704-084830


salaamdev:
Good catch on convertTask. The old .day == today.day check labeled anything due on the same day-of-month as "today", so an overdue task from July 9 would sit in Today on August 9. The full year/month/day compare with .toLocal() fixes that properly, and the null guard on dueDate was overdue too. I was about to file a PR for this myself, glad you covered it.

bgrassy:
> Good catch on convertTask. The old .day == today.day check labeled anything due on the same day-of-month as "today", so an overdue task from July 9 would sit in Today on August 9. The full year/month/day compare with .toLocal() fixes that properly, and the null guard on dueDate was overdue too. I was about to file a PR for this myself, glad you covered it.

Yeah, those fixes caught a good amount of issues surrounding the due date behavior for tasks in the widget. I've been using my version for a while and it's been generally solid, but there are definitely some quirks surrounding offline behavior and formatting of dates that could use fixing in a future PR.

leschbenedikt (approved):

PR #292, opened 2026-07-12
Add German localization strings to app_de.arb

skar-cgn:
Create German translation (with help of AI)


kolaente:
Translations should go through Crowdin, as per the translation guidelines. https://vikunja.io/docs/translations/

Issue #57, opened 2024-01-29
Subtasks are not indented


BeatLink:
On the home page and in the projects page, subtasks are treated as top level tasks instead of being indented like in the web app

PR #294, opened 2026-07-15
Addressing subtask indentation feature request

AsbestosLampshade:
The subtask request #57 should be fixed with this change, tested on an android emulator, g_pixel_34. 
Subtasks seem  to generate fine.
image
image

Please do let me know if any changes are required.


AsbestosLampshade:
Hi @leschbenedikt, could you help push the review the code and push it for CI when available?

lib/data/models/task_dto.dart
```diff
@@ -188,4 +184,32 @@ class TaskDto extends Dto<Task> {
     bucketId: b.bucketId,
     createdBy: b.createdBy != null ? UserDto.fromDomain(b.createdBy!) : null,
   );
+
+  static int? _extractParentTaskId(Map<String, dynamic> json) {
```
leschbenedikt, line 188: If I understand that correctly 'related_tasks' is the new api and it was 'parent_task' before.

We don't want to support outdated api from the pre 1.0 release in the app. So please use only the newest one.
For future changes we will have api versioning.

leschbenedikt (commented):
Thanks for you work.

Unfortunately this does not work correctly at the moment.

I've created a task with a sub task which has a sub task itself like this.

Parent Task -> Sub Task -> Sub sub Task


On the web page this will be displayed like this in overview and upcoming:

Image

In the project page, it will be displayed like this:

Image



In the home page of the android app the sub sub taks is missing.
Also they are indented which differs from the web page:

Image

In the project page the indentation would be correct but also sub sub task is missing:

Image

AsbestosLampshade:
image -- Home
Multi-layer tasks are accommodated for, thanks for the pointer, never used >2, so I never noticed.
Let me know if the indentation matches expectation here
image -- Within Projects

image
Dropped parent_task_id as source here.

Thanks a lot for your comments @leschbenedikt 
Do let me know what changes you think I should add here.

leschbenedikt:
Hey,
thanks for the changes. Unfortunately not everything seems to be fixed.

The sub sub tasks is still not visible at all for me.

leschbenedikt (changes requested):

PR #301, opened 2026-08-07
feat: switch android to okhttp, add mtls support

### LINKED ISSUE #147: Allow using client certificates
The app does not seem to allow using client certificate for setting up the connection to the server.
When connecting to a server tgat requires a client certificate. The app does not prompt the user to choose one.

nijave:
AI assisted

Closes #147 

Might need a bit of cleanup, don't have much Android/Dart experience


android/app/build.gradle
```diff
@@ -47,7 +47,7 @@ android {
 
     defaultConfig {
         applicationId "io.vikunja.app"
-        minSdkVersion flutter.minSdkVersion
+        minSdkVersion 24
```
nijave, line 50: 21->24 drops android 5/6 suport

leschbenedikt:
I really appreciate the work and it really looks like it was a lot.

But I have some major concerns about this PR.
The biggest is that this is all based on an library that is still in experimental state and has only one release that is 2 years old.
I really don't feel like having a dependency like that in the code, especially in the core of it.

And also this will only work for Android and not iOS where this app is meant for both

nijave:
> I really appreciate the work and it really looks like it was a lot.
> 
> But I have some major concerns about this PR. The biggest is that this is all based on an library that is still in experimental state and has only one release that is 2 years old. I really don't feel like having a dependency like that in the code, especially in the core of it.
> 
> And also this will only work for Android and not iOS where this app is meant for both

Yes, that's fair. Cronet doesn't support system managed client certs on Android. I could only find a single issue requesting it on the issue tracker and it's had no feedback and was opened a year ago, unfortunately. The only option for that library is to add functionality to pick a file from storage and have the app create a security context although I haven't tested that route since I was hoping to use the system certificate manager (where you can use a certificate imported in settings that gets securely stored in TEE). I think Immich does the cronet route.

If you're okay with okhttp route, I can poke the dart maintainers about it since it's been in tree I think 4 years without a status update. I guess another option is adding Kotlin here to wrap okhttp directly.

I think SDK min version bump came with ok_http dart package although 24 is still quite old (I think 2016). Maybe Sentry has device stats to see if there's any usage?

As for iOS, that's quite a stretch for me. The only Mac I have is my company MBP I'd prefer not to use. The Android version I've been using on my personal device since the PR was opened with the mTLS stack here https://github.com/nijave/vmubtkube-a/blob/main/proxy_vikunja_apps_somemissing_info.yaml#L25

On iOS, I think it would need switched to cupertino_http based on my reading last night and it has a slightly different in app cert import process. I guess on iOS there's no global cert store apps have access to.

Presumably Immich has some prior art here as well but I know much less on the iOS side.

If you have some time I'd appreciate feedback on the direction like you'd like to see this go given the above details and I can make some changes. I can also break into smaller PRs as there's a fair bit of wiring with http client swap + cert selection machinery and error handling.

leschbenedikt:
I'm not against okhttp as library but I'm at the moment not sure if that's the right way or more a bonus for users who want to use the system certificate manager.

I also don't think iOs does have something like this so we still need the solution of picking a file from storage so we can make it work there. And that would work on Android too. I'm on a mac so I can help out testing that.
After that we already would have a working solution and the okhttp route is just a cherry on top with some downsides.

For example, your change removed the background downloader for attachments. So users can't leave the page or close the app while downloading.

Also leaving this feature out of scope, I think cronet is the better path for this application.

In general I think we need to find a middle ground and answer some questions before.
Like, do we need the system certificate manager or is it just a nice to have?
What exactly do we need to make it work on iOs?
Can we get background download working with this solution or find something similar?

Also to notice is that immich does indeed use ok_http but not the official version like in this PR but
a fork of it. I bet there is a reason behind it.
Maybe we could get some inspiration by looking at what they did

PR #308, opened 2026-09-02
feat: add search for tasks and projects

VARWA:
## Summary
- add a reusable searchable app bar and wire it into home tasks, project list, and project task list views
- implement server-backed search for tasks and projects while preserving existing filters and pagination behavior
- refine search UX behavior (focus, keyboard dismissal, back navigation, empty states) and add tests for new search helpers and UI states

## Test plan
- [x] flutter test test/search_filter_test.dart test/presentation/manager/projects_controller_test.dart test/presentation/pages/task/task_list_page_test.dart test/presentation/pages/project/project_list_page_test.dart test/presentation/widgets/project/project_task_list_test.dart test/localization_test.dart
- [x] Manual verification on mobile emulator: search open/close behavior, keyboard focus, and result updates

Made with [Cursor](https://cursor.com)


lib/presentation/manager/project_controller.dart
```diff
@@ -14,6 +15,9 @@ part 'project_controller.g.dart';
 
 @riverpod
 class ProjectController extends _$ProjectController with PaginationMixin<Task> {
+  int _searchEpoch = 0;
+  String _searchQuery = '';
```
leschbenedikt: Why do we need that as fields?
We could just check against the model

lib/presentation/manager/project_controller.dart
leschbenedikt, line 62: If a request of loadNextPage is running while we start a search the result appends values that possibly don't match.

We should guard this call too

lib/presentation/manager/task_page_controller.dart
leschbenedikt, line 62: Needs to be guarded too like in project controller

lib/presentation/manager/task_page_controller.dart
```diff
@@ -14,6 +15,9 @@ part 'task_page_controller.g.dart';
 @riverpod
 class TaskPageController extends _$TaskPageController
     with PaginationMixin<Task> {
+  int _searchEpoch = 0;
+  String _searchQuery = '';
```
leschbenedikt: Use value from model

lib/presentation/manager/projects_controller.dart
```diff
@@ -10,6 +10,9 @@ part 'projects_controller.g.dart';
 @riverpod
 class ProjectsController extends _$ProjectsController
     with PaginationMixin<Project> {
+  int _searchEpoch = 0;
+  String _searchQuery = '';
```
leschbenedikt: Use value from model

lib/presentation/manager/projects_controller.dart
leschbenedikt, line 94: Needs guard to, same as in project controller

leschbenedikt (changes requested):
Thanks for you contribution. Found a few thing that should be fixed. But in general I like the idea of having a search

VARWA:
Thanks for the review. Both points are addressed: search now uses the value from the model instead of extra controller fields, and loadNextPage is guarded in all three controllers so an in-flight page request cannot append results after a search has started.

PR #309, opened 2026-09-02
feat: improve task creation options

VARWA:
## Summary
- allow selecting the destination project when creating a task from the overview or quick-add flow
- replace the expanded due-date chips with a compact, stable selector and custom date/time picker
- add priority selection with colors matching the existing task priority indicators
- preserve the current project in project and Kanban task creation flows

## Test plan
- [x] `flutter test test/presentation/widgets/task/add_task_dialog_test.dart`
- [x] `flutter test test/presentation/pages/task/task_list_page_test.dart test/presentation/pages/project/project_list_page_test.dart test/presentation/widgets/project/project_task_list_test.dart`
- [x] Manually verify project, due date, and priority selection on a mobile device

Made with [Cursor](https://cursor.com)


leschbenedikt (changes requested):
I like the idea of that change. 
But I think we can improve the design a little.
In my opinion it looks a little outdated right now.

Maybe we could stick more to the design as it was and add the new feature to it

VARWA:
> I like the idea of that change. But I think we can improve the design a little. In my opinion it looks a little outdated right now.
> 
> Maybe we could stick more to the design as it was and add the new feature to it

Thanks for the feedback. I restored the original due-date chips and underline inputs, and added project + priority in the same style as the task edit screen. I also made the dialog a bit more compact and scrollable when the keyboard is open.

lib/presentation/widgets/task/add_task_dialog.dart
```diff
@@ -18,133 +35,247 @@ class AddTaskDialogState extends State<AddTaskDialog> {
...
+  void _onTitleChanged() {
+    setState(() {});
   }
 
   @override
   Widget build(BuildContext context) {
     var dateTime = DateTime.now();
+    final l10n = AppLocalizations.of(context);
+
+    final remainingHeight =
```
leschbenedikt, line 69: Calculating sizes is most often not a good idea.
In this case a small device a landscape mode can trigger an error.

Image

lib/presentation/pages/home_page.dart
```diff
@@ -124,39 +125,59 @@ class HomePageState extends ConsumerState<HomePage> {
   }
 
   Future<dynamic> showAddItemDialog(String? title) async {
-    var response = await ref.read(userRepositoryProvider).getCurrentUser();
+    var userResponse = await ref.read(userRepositoryProvider).getCurrentUser();
+    var projectsResponse = await ref.read(projectRepositoryProvider).getAll();
```
leschbenedikt, line 129: Doesn't load all projects if you page more than 50.
Also only show projects with id > 0. Others are Favorites and saved filters.

lib/presentation/widgets/task/add_task_dialog.dart
```diff
@@ -18,133 +35,247 @@ class AddTaskDialogState extends State<AddTaskDialog> {
...
+  int priority = 0;
 
   @override
   void initState() {
     super.initState();
 
     var title = widget.title;
     if (title != null) {
-      textController.text = title;
+      textController.text = _capitalizeFirst(title);
```
leschbenedikt, line 47: Why are we capitalizing the first letter of the title?
Especially without the user knowing.

The task should have the title the user entered

lib/presentation/pages/task/task_list_page.dart
```diff
@@ -44,14 +46,18 @@ class TaskListPage extends ConsumerWidget {
             ),
           ),
           floatingActionButton: FloatingActionButton(
-            onPressed: () {
-              if (model.defaultProjectId == 0) {
+            onPressed: () async {
+              final projects = await _loadProjects(ref);
```
leschbenedikt, line 50: After the button press we are now waiting on the network call.
On a slow connection the user will get no feedback just doesn't see the dialog open.

We should open the dialog directly, the project should be already available anyway

lib/presentation/pages/home_page.dart
```diff
@@ -124,39 +125,59 @@ class HomePageState extends ConsumerState<HomePage> {
...
-        return Future.value();
+        return;
       }
+
+      var defaultProjectId = userResponse
+          .toSuccess()
+          .body
+          .settings
+          ?.defaultProjectId;
+      final selectedProjectId =
```
leschbenedikt, line 151: Different default project id logic than in task_list_page

lib/presentation/widgets/task/add_task_dialog.dart
```diff
@@ -1,14 +1,31 @@
...
+import 'package:vikunja_app/domain/entities/project.dart';
 import 'package:vikunja_app/l10n/gen/app_localizations.dart';
+import 'package:vikunja_app/presentation/widgets/date_time_field.dart';
 
 class AddTaskDialog extends StatefulWidget {
-  final void Function(String title, DateTime? dueDate) onAddTask;
+  final void Function(
+    String title,
+    DateTime? dueDate,
+    int? projectId,
```
leschbenedikt, line 14: It makes no sense to have a nullable project Id here. A project needs to be selected.
Also every caller direct replaces it with the initial value if null

lib/presentation/pages/project/project_detail_page.dart
```diff
@@ -176,6 +184,8 @@ class ProjectPageState extends ConsumerState<ProjectDetailPage> {
     Project project,
```
leschbenedikt, line 184: Why do we have project and projectId here

lib/presentation/widgets/project/kanban/bucket_column.dart
leschbenedikt, line 371: We pass projectId from the dialog but use project from the widget here.
Should always be the same but very confusing

lib/presentation/widgets/task/add_task_dialog.dart
```diff
@@ -160,4 +291,20 @@ class AddTaskDialogState extends State<AddTaskDialog> {
       },
     );
   }
+
+  Color _priorityColor(BuildContext context, int value) {
```
leschbenedikt, line 295: Duplicate of PriorityBatch.getBackgroundColor

lib/presentation/widgets/task/add_task_dialog.dart
```diff
@@ -18,133 +35,247 @@ class AddTaskDialogState extends State<AddTaskDialog> {
...
-              ),
-              if (dateTime.hour < 21)
-                taskDueList(
-                  AppLocalizations.of(context).dueOptionToday,
-                  NewTaskDue.today,
+              TextField(
+                keyboardType: TextInputType.text,
+                textCapitalization: TextCapitalization.sentences,
+                textInputAction: TextInputAction.done,
+                maxLines: 1,
```
leschbenedikt, line 94: Having multi line support here was intentional

lib/presentation/widgets/task/add_task_dialog.dart
```diff
@@ -18,133 +35,247 @@ class AddTaskDialogState extends State<AddTaskDialog> {
...
+                    isDense: true,
+                    icon: Icon(
+                      key: ValueKey('priority-icon-$priority'),
+                      priority == 0 ? Icons.flag_outlined : Icons.flag,
+                      color: _priorityColor(context, priority),
+                    ),
+                    labelText: l10n.priority,
+                    border: InputBorder.none,
+                  ),
+                  selectedItemBuilder: (context) => List.generate(
```
leschbenedikt, line 200: Deprecated. Use initialValue

lib/presentation/widgets/task/add_task_dialog.dart
```diff
@@ -18,133 +35,247 @@ class AddTaskDialogState extends State<AddTaskDialog> {
...
+  }
+
+  @override
+  void dispose() {
+    textController.removeListener(_onTitleChanged);
+    textController.dispose();
+    super.dispose();
+  }
+
+  void _onTitleChanged() {
```
leschbenedikt, line 60: Why does every keystroke rebuild the complete dialog?

test/presentation/widgets/task/add_task_dialog_test.dart
```diff
@@ -0,0 +1,63 @@
...
+
+    await tester.tap(find.text('Unset'));
+    await tester.pumpAndSettle();
+    await tester.tap(find.text('High').last);
+    await tester.pumpAndSettle();
+
+    final priorityIcon = tester.widget<Icon>(
+      find.byKey(const ValueKey('priority-icon-3')),
+    );
+    expect(priorityIcon.color, Colors.red);
```
leschbenedikt, line 55: Why are we testing for a backup color?

leschbenedikt (changes requested):

PR #315, opened 2026-09-18
test: add layered test suite

leschbenedikt:
Add 1078 tests covering 86% of non-generated lines, all passing against the current app code. Each behaviour is tested once, at the lowest layer that owns it; test/README.md documents the layers, the shared helpers and what is deliberately not covered. UI tests render at phone width.

- Shared helpers in test/helpers: entity builders, JSON fixtures, recording fakes for every repository and controller, platform-channel stubs, an HttpClient stand-in and a phone-sized pumpApp.
- `make coverage` and CI report coverage per area, excluding generated code, and fail below 80%.
- The add-task dialog, DueDateCard and convertTask read the time through clock.now() instead of DateTime.now(), so tests can pin "now" rather than race the real clock. No behaviour changes.

Tests for the defects this suite found come with their fixes in a follow-up change.


kolaente:
What's the reasoning behind the tests? What did you prompt?

And please split these up in more commits, 15k lines is too much for one commit

leschbenedikt:
The reasoning was to be more robust to changes to hopefully have less regression.

I have a lot of changes in the pipeline and just want to be more sure it doesn't break things.

The prompt was just something like "Test coverage of 80%. Every user interaction. No overlaps/duplications"
And then a loop of 10 reviews from a different session feeding it to the initial and so on..

I just did it in one go as it doesn't touch the code itself. But I can split it if you want. 
How do you want me to split it?

